PR #4, opened 2019-06-06
`list` command for Windows

yuchong-pan:
1. Implemented `list` command for Windows, using data fetched from the registry.
    * The registry key used is `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall` so that both 32-bit and 64-bit .NET Core SDKs are displayed. We may change this if we want to display 64-bit .NET Core SDKs only. Also `--list-sdks` only displays 64-bit SDKs.
    * The display string is subject to change. Currently it just displays the `DisplayName` from the registry, e.g., "Microsoft .NET Core SDK 2.2.300 (x64)". However, the `--list-sdks` displays "2.2.300 [C:\Program Files\dotnet\sdk]".
2. Added unit tests for essential methods.


src/dotnet-uninstall/Program.cs
```diff
@@ -2,14 +2,49 @@
...
+            rootCommand.Add(uninstallCommand);
+
+            listCommand.Handler = CommandHandler.Create(() =>
+            {
+                ExecuteListCommand();
+            });
+
+            uninstallCommand.Handler = CommandHandler.Create(() =>
+            {
+                Console.WriteLine("invoked uninstall subcommand"); // TODO: implement uninstall
```
wli3: minimize todo. this can track in "issues". And we can just point "not support" for now

test/dotnet-uninstall.Tests/Windows/RegistryManagerTests.cs
```diff
@@ -0,0 +1,252 @@
...
+        {
+            RegistryManager.IsDotNetCoreSdkPublisher(MicrosoftCorporationPublisher)
+                .Should().BeTrue();
+        }
+
+        [Fact]
+        public void TestIsDotNetCoreSdkDisplayNameNull()
+        {
+            RegistryManager.IsDotNetCoreSdkDisplayName(null)
+                .Should().BeFalse();
```
wli3: you can use `[Theory]` to condense these like https://github.com/dotnet/sdk/blob/f0a1246be1ebe11a88097ac3dd5736f0ad15b855/src/Tests/Microsoft.NET.Build.Tests/GivenThatWeWantBuildsToBeIncremental.cs#L23

src/dotnet-uninstall/Program.cs
```diff
@@ -2,14 +2,49 @@
...
+        }
+
+        private static void ExecuteListCommand()
+        {
+            if (RunningOnWindows)
+            {
+                Windows.ListCommand.Execute();
+            }
+
+            // TODO: implement list for macOS and linux
```
wli3: same here

src/dotnet-uninstall/Program.cs
yuchong-pan: done

src/dotnet-uninstall/Program.cs
yuchong-pan: done

test/dotnet-uninstall.Tests/Windows/RegistryManagerTests.cs
yuchong-pan: done

Issue #7, opened 2019-06-06
Add localization infrastructure


yuchong-pan:

wli3:
reference this https://github.com/dotnet/cli-lab/pull/10

Issue #1, opened 2019-05-29
Design Description - .NET Core installation cleanup tool


leecow:
## Problem

From the beginning, .NET Core was optimized for side-by-side installations to solve a number of problems inherent in the .NET Framework installation and deployment models. Consequently, 1.x and 2.x installers for Windows and macOS would install successive updates to versioned sub-directories. Because of a desire to enable easy fallback to previously installed versions, the installers did not provide functionality which removed previous patch versions as updates were installed. This resulted in systems with a high number of patch versions installed. The problem is particularly acute in Visual Studio scenarios where users may have opted into Preview channels which see a significantly higher number of updates.

Strongly negative feedback from internal and external customers began increasing in the 2.1 and 2.2 development cycles.

Currently, the only options for removing unwanted versions of the .NET Core Runtimes and SDKs is to manually walk through Add/Remove Programs (Windows) or manual deletion of directories (macOS). This issue does not impact Linux installations as the update functionality was implemented for the 2.1 release.

## Proposal

The Windows installer for 3.0 has been updated to remove the previous patch version after a new release has been installed. Similar updates will be made to the Mac installer though this may not be possible for the 3.0 release.

To assist customers with pre-3.0 installation cleanup, a guided tool will be provided to enable the controlled clean up of a system such that only the desired versions of the Runtime and SDKs remain. Control input will be through a combination of the command line and interactive input.

## dotnet-uninstall

Verb      | Argument or option            | Help text/Description                                                                               | Notes                                                      |
|-----------|-------------------------------|---------------------------------------------------------------------------------------------------|------------------------------------------------------------|
| list      |                               | List the .NET Core SDKs that are installed globally on the machine                                |  This is the same as  dotnet --list-sdks                   |
| uninstall | <string[] versions>           | Remove the .NET Core SDKs specified                                                               |                                                            |
| uninstall | --all                         | Remove all .NET Core SDKs installed globally on the machine                                       |                                                            |
| uninstall | --all-lower-patches           | Remove .NET Core SDKs that have been replaced by higher patches                                   |                                                            |
| uninstall | --all-but-latest              | Remove all .NET Core SDKs, except the highest version                                             |                                                            |
| uninstall | --all-but <string[] versions> | Remove all .NET Core SDKs, except those listed                                                    | Not logical with `uninstall   <string[] versions>          |
| uninstall | --all-below <string version>  | Remove all .NET Core SDKs below  specified version. The specified version will remain             |                                                            |
[...]

PR #9, opened 2019-06-10
Platform-independent code for .NET Core installation clean-up tool

yuchong-pan:
1. Refactored code structure to maximize code reuse. All platform-independent code is moved to the `Shared` namespace. Platform-independent code is covered by unit tests.
2. All user-faced strings are in `Messages.cs`, to be localized. See Issue #7 and Pull Request #10. Note that these user-faced strings are all subject to discussion and change.
3. _Temporarily_, I use `--sdk` and `--runtime` to distinguish SDK and runtimes, though these options are not currently in the design description. During an offline discussion, we agreed that options like these will be added to the syntax, and is to be discussed. For certain options, I also require exactly one of `--sdk` and `--runtime` needs to be specified, otherwise I don't know whether the specified version is for SDKs or for Runtimes; e.g., `dotnet uninstall 2.0.5`. For other options, not specifying a bundle type means the command applies to all bundle types.

UI issues to consider (with responses from @KathleenDollard):
1. Shall we do something like a `--do-it` flag or a `y/n` prompt, as described in Issue https://github.com/dotnet/cli-lab/issues/1#issuecomment-497222636?
_Response_: We need to ask for opinions from software engineers on the team. Suspend the implementation.

2. Shall we add an option to distinguish x64 and x86 platforms?
_Response_: Yes, `--x64` and `--x86` for now. The default is both `--x64` and `--x86`. We allow users to specify both.

3. Shall we add an option to distinguish SDK, Runtimes and other bundles (e.g. ASP.NET)? For certain options, shall we required _exactly_ one of the bundle types needs to be specified? (See 3 above.)
_Response_: Yes, `--sdk` and `--runtime` for now. We do _not_ allow users to specify both, and we don't provide a default option. An exception is the `list` command, where we allow users to specify both and the default option is both.

4. Is the command line help information generated by the `System.CommandLine` package confusing?
    * `dotnet-uninstall [options] <SDK-VERSIONS> [command]`
    * But we only allow _exactly one_ of `[options]`, `<SDK-VERSIONS>`, `[command]` to be specified.

   _Response_: Post an issue on `System.CommandLine`.

5. Does the latest version in the `--all-but-latest` option include the preview versions?
_Response_: Yes.

6. Since we take command arguments (e.g., `dotnet uninstall 2.2.300`), we will regard unknown options as versions (e.g., `dotnet uninstall --unknown`) and users will get error message "The specified version --unknown is not valid." Is this confusing?
_Response_: Post an issue on `System.CommandLine`.

7. What should the error message be when the user types `dotnet uninstall` only?
_Response_: Something like "Required argument missing".

Note: the answers to the above questions are all subject to change based on feedback.

What's _NOT_ in the pull request?
1. The localization infrastructure.
2. The macOS version of detecting installed .NET Core SDKs, Runtimes and other bundle types.
3. Detection of ASP.NET bundles and other bundle types (e.g. .NET Core Hosting). To be discussed.
4. The actual uninstallation execution and system permission handling for both platforms.
5. Different verbosity levels. To be discussed.
6. @KathleenDollard's review on user-faced strings.
7. I plan to make changes to the implementation of the syntax according to @KathleenDollard's replies after this PR, as this is already a big PR.

[This pull request is ready to be reviewed and merged.]


src/dotnet-uninstall/Shared/Commands/UninstallCommandExec.cs
```diff
@@ -0,0 +1,89 @@
...
+            }
+
+            return specifiedOption;
+        }
+
+        private static void ExecuteUninstall(IEnumerable<ISdkInfo> sdks)
+        {
+            foreach (var sdk in sdks)
+            {
+                // TODO: replace this
```
wli3: This is how CLI does it https://github.com/dotnet/cli/blob/95c6eff6daa1a69f29c42b2d405400ad44bdec91/src/Microsoft.DotNet.Cli.Utils/AnsiColorExtensions.cs#L6

src/dotnet-uninstall/Shared/SdkInfo/SdkVersion.cs
```diff
@@ -0,0 +1,89 @@
+﻿using System;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Shared.SdkInfo
+{
+    class SdkVersion : IEquatable<SdkVersion>, IComparable<SdkVersion>, IComparable
```
wli3: I like this class a lot.

nit: add explicit internal

yuchong-pan: I'm just wondering why explicit `internal`s are preferred, cus I just removed all `internal`s to make code clearer.

yuchong-pan: Will fix that. [Discussed]

yuchong-pan: done.

src/dotnet-uninstall/Shared/Commands/UninstallCommandExec.cs
yuchong-pan: [Discussed, stay with `Console.ForegroundColor`]

src/dotnet-uninstall/Messages.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace Microsoft.DotNet.Tools.Uninstall
+{
+    internal static class Messages
```
livarcocc: Should this be localizablestrings.resx, preparing it for localization? Maybe file an issue do make the localizablestrings change later on, given this is already a big PR.

src/dotnet-uninstall/Shared/BundleInfo/BundleVersion.cs
```diff
@@ -0,0 +1,109 @@
...
+
+        public override bool Equals(object obj)
+        {
+            return Equals(obj as BundleVersion);
+        }
+
+        public bool Equals(BundleVersion other)
+        {
+            return other != null &&
+                   Major == other.Major &&
```
livarcocc: Just checking _versionString is not enough, after the null check?

src/dotnet-uninstall/Shared/BundleInfo/BundleVersion.cs
```diff
@@ -0,0 +1,109 @@
...
+                   Minor == other.Minor &&
+                   Patch == other.Patch &&
+                   Preview == other.Preview;
+        }
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(Major, Minor, Patch, Preview);
```
livarcocc: Same here. Should we just hash the string?

src/dotnet-uninstall/Messages.cs
yuchong-pan: Yes, @wli3 said that he would add the localization infrastructure. It's already in Issue #7.

src/dotnet-uninstall/Shared/BundleInfo/BundleVersion.cs
yuchong-pan: I don't think so, because different strings may correspond to the same version (there are different version string formats for SDK and Runtime). So even if two `_versionString`s are different, the versions they correspond to are the same, e.g. `3.0.100 - preview5` and `3.0.1 Preview 5`.

src/dotnet-uninstall/Shared/BundleInfo/BundleVersion.cs
yuchong-pan: Same as above.

src/dotnet-uninstall/Shared/Commands/UninstallCommandExec.cs
wli3: nit: remove this, it is done

src/dotnet-uninstall/Shared/Configs/OptionFilterers.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System.Collections.Generic;
+using System.CommandLine;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Filterers;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Shared.Configs
+{
+    internal static class OptionFilterers
```
wli3: OptionFilterersMap?

src/dotnet-uninstall/Shared/Exceptions/DotNetUninstallExceptions.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
+using System.CommandLine;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Shared.Exceptions
+{
+    internal abstract class DotNetUninstallException : Exception
```
wli3: nit: make them different files. just how we usually do it

wli3:
Great job overall!

src/dotnet-uninstall/Shared/Configs/OptionFilterers.cs
yuchong-pan: For the `TODO` in `UninstallCommandExec`, it's indeed not done. The console print is just a placeholder and it should be replaced by the actual uninstallation execution, which is in another PR.

src/dotnet-uninstall/Shared/Configs/CommandLineConfigs.cs
```diff
@@ -0,0 +1,174 @@
...
+            UninstallAllLowerPatchesOption,
+            UninstallAllButLatestOption,
+            UninstallAllButOption,
+            UninstallAllBelowOption,
+            UninstallAllPreviewsOption,
+            UninstallAllPreviewsButLatestOption,
+            UninstallMajorMinorOption,
+        };
+
+        public static readonly IEnumerable<Option> AuxOptions = new Option[]
```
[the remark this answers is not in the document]
yuchong-pan: `Aux` is for auxiliary

test/dotnet-uninstall.Tests/Shared/Filterers/MajorMinorOptionFiltererTests.cs
```diff
@@ -0,0 +1,120 @@
+﻿using System.Collections.Generic;
+using System.CommandLine;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Configs;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Exceptions;
+using Xunit;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Tests.Shared.Filterers
+{
+    public class MajorMinorOptionFiltererTests : FiltererTests
```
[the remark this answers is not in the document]
yuchong-pan: I think the `--major-minor` option is quite clear, as it just requires something like `2.2`. I think what has not been decided is the other options. However I will double-check with @KathleenDollard.

test/dotnet-uninstall.Tests/Shared/Filterers/FiltererTests.cs
```diff
@@ -0,0 +1,109 @@
...
+using FluentAssertions;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Configs;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Exceptions;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Filterers;
+using Xunit;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Tests.Shared.Filterers
+{
+    public abstract class FiltererTests
```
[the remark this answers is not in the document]
yuchong-pan: [Discussed offline; the inheritance will be kept in this case.]

src/dotnet-uninstall/Shared/Configs/OptionFilterers.cs
yuchong-pan: [Discussed offline; keep `OptionFilterers`]

src/dotnet-uninstall/Shared/Exceptions/DotNetUninstallExceptions.cs
yuchong-pan: Done.

test/dotnet-uninstall.Tests/Shared/Filterers/FiltererTests.cs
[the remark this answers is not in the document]
yuchong-pan: [Discussed offline; otherwise the filterer names will be ridiculously long, e.g. `AllPreviewsButLatestOptionBundleFilterer`]

test/dotnet-uninstall.Tests/Shared/Filterers/MajorMinorOptionFiltererTests.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        internal override Option Option => CommandLineConfigs.UninstallMajorMinorOption;
+        internal override string DefaultArgValue => "2.2";
+        internal override bool TestBundleTypeNotSpecifiedException => false;
+
+        public static IEnumerable<object[]> GetDataForTestFiltererGood()
```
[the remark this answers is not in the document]
yuchong-pan: I still hold that such a name would be enough to communicate the meaning of the data, because the parameter names are in the signature of the actual testing method.

See https://github.com/dotnet/cli/blob/6da8a92712d114af8131c2683c0c0f78436f6d80/test/Microsoft.DotNet.Cli.Utils.Tests/StreamForwarderTests.cs.

src/dotnet-uninstall/Shared/Configs/OptionFilterers.cs
yuchong-pan: Replacing `TODO` with `NotImplementedException` is done. I can't reply to and resolve that comment.

test/dotnet-uninstall.Tests/Shared/Filterers/FiltererTests.cs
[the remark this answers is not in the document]
wli3: inheritance is fine. But you need to find some way to make input, and expectation stand out. This is something I will block on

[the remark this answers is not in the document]
wli3: I think we missed something. I don't think it is "ridiculously long". I think it is a better name since ["honest and complete" is better than "honest"](http://arlobelshee.github.io/Presentations/legacy_code/7_stages_of_naming.png). But I won't block on this issue

test/dotnet-uninstall.Tests/Shared/Filterers/MajorMinorOptionFiltererTests.cs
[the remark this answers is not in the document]
wli3: In ForwardingTheoryVariations, at least before the comma it is the input and after the comma it is the expectation. And in this PR case, it is hard to find what is the input

wli3:
The only blocking issue is https://github.com/dotnet/cli-lab/pull/9#discussion_r294615482 .

test/dotnet-uninstall.Tests/Shared/Filterers/MajorMinorOptionFiltererTests.cs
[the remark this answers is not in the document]
yuchong-pan: Done.

test/dotnet-uninstall.Tests/Shared/Filterers/FiltererTests.cs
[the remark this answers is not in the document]
yuchong-pan: Done.

src/dotnet-uninstall/Shared/BundleInfo/RuntimeVersion.cs
```diff
@@ -0,0 +1,69 @@
+﻿using System;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Utils;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo
+{
+    internal class RuntimeVersion : BundleVersion, IEquatable<RuntimeVersion>, IComparable, IComparable<RuntimeVersion>
```
livarcocc: Can you help me understand the difference between bundleversion and runtimeversion here?

src/dotnet-uninstall/Shared/Configs/CommandLineConfigs.cs
```diff
@@ -0,0 +1,174 @@
...
+            {
+                UninstallRootCommand.AddOption(option);
+                ListCommand.Add(option);
+            }
+
+            ListCommand.Handler = CommandHandler.Create(ExceptionHandler.HandleException(() => ListCommandExec.Execute()));
+            UninstallRootCommand.Handler = CommandHandler.Create(ExceptionHandler.HandleException(() => UninstallCommandExec.Execute()));
+        }
+
+        public static Option GetUninstallMainOptions(this CommandResult commandResult)
```
livarcocc: NIT: Should this say GetUninstallMainOption, singular, given that having more than one specified will throw an exception?

src/dotnet-uninstall/Shared/Utils/Regexes.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System.Text.RegularExpressions;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Shared.Utils
+{
+    internal static class Regexes
```
livarcocc: This makes me want to run for the hills.

src/dotnet-uninstall/Shared/Configs/CommandLineConfigs.cs
yuchong-pan: Thanks for the catch!! It is indeed a typo and I intended to call it `GetUninstallMainOption`

yuchong-pan: Done.

src/dotnet-uninstall/Shared/BundleInfo/RuntimeVersion.cs
yuchong-pan: `BundleVersion` is an abstract class and cannot be instantiated. It is the base class of both `SdkVersion` and `RuntimeVersion`. This way we can have for instance `List<BundleVersion>` so that both `SdkVersion` and `RuntimeVersion` can be included.

The reason I would like to distinguish SDK versions and Runtime versions is that SDK versions are of the form `{major}.{minor}.{sdkMinor}{patch}` whereas Runtime versions are semvers. Hence SDK versions require an additional field.

See https://docs.microsoft.com/en-us/dotnet/core/versions/.

src/dotnet-uninstall/Shared/Utils/Regexes.cs
yuchong-pan: I agree. This class is kind of messy. But if we don't do this then we would have duplicate regexes spread in the entire project and updating a regex requires updating those copies, which I think is very hard to maintain.

Issue #8, opened 2019-06-07
Add Version.Details.xml to finish publish


wli3:
Failed build
https://dev.azure.com/dnceng/internal/_build/results?buildId=215182

PR #11, opened 2019-06-13
Add Version.Details.xml

wli3:
fix https://github.com/dotnet/cli-lab/issues/8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -141,7 +141,6 @@ node_modules/
 *.Tests/lib/
 *.userprefs
 *.vsp
-*.xml
 .idea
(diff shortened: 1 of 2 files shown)


wli3:
Once this is merge you can find built package at nuget feed https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json you can use dotnet tool install dotnet-uninstall --add-source https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json --version VERSION

Or could use https://github.com/NuGetPackageExplorer/NuGetPackageExplorer to see it

PR #12, opened 2019-06-17
Uninstallation command execution for Windows

yuchong-pan:
List of UX issues to consider:
1. Currently, users will be prompted for "User Account Control" confirmation for _each_ of the .NET Core bundles to be uninstalled. This can be annoying for users. From the user experience perspective, shall we ask for the administrator permission when the tool starts? Alternatively, users can run `cmd.exe` as an administrator and they won't be prompted for confirmations again. I'm not sure which is the way that Windows console apps usually take.
_Response_: We need to be able to elevate privileges only once, as the approach of running `cmd.exe` as administrator has security concerns.

2. Is it acceptable that the `cmd` window on which the clean-up tool is running will pop up a new `cmd` window if the current `cmd` does not have administrator privileges? This, however, is a trade-off so that the user will be able to uninstall _all_ .NET Core bundles on their machine. In addition, otherwise we do not have a way to redirect the output of the new process so far because of the administrator privileges.

3. The users will not be able to press Control-C to cancel the uninstallation(s). Is this acceptable? If not, this requires a bit of technical exploration.


rainersigwald:
I'd recommend opting for a single elevation if possible. One option would be to detect the non-elevated case and launch yourself again, elevated, with an argument indicating not to check.

KathleenDollard:
1) We should probably involve the security folks at this point. @wli3 can you make the connection. 
2) I'm fine if we have some rough edges like this
3) I'm concerned about blocking Ctl-C on a destructive command. I'd like us to look for other approaches to see what trade offs we have to make.

yuchong-pan:
> 1. We should probably involve the security folks at this point. @wli3 can you make the connection.
> 2. I'm fine if we have some rough edges like this
> 3. I'm concerned about blocking Ctl-C on a destructive command. I'd like us to look for other approaches to see what trade offs we have to make.

Thank you @KathleenDollard for your reply. For 3, would it be okay if we have another key combination in place of Ctrl-C for user cancellation?

src/dotnet-uninstall/Windows/ProcessHandler.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            if (!IsAdmin())
+            {
+                RunAsAdmin();
+                return;
+            }
+
+            foreach (var bundle in bundles.ToList().AsReadOnly())
+            {
+                var args = ParseCommand(bundle.UninstallCommand, out var argc);
```
wli3: what is "argc"

src/dotnet-uninstall/Windows/ProcessHandler.cs
```diff
@@ -0,0 +1,123 @@
...
+
+            if (!process.Start())
+            {
+                throw new ElevationFailedException();
+            }
+
+            Environment.Exit(0);
+        }
+
+        private static IEnumerable<string> ParseCommand(string command, out int argc)
```
wli3: might be hard to convey that in code or tests, so we need some comment for why you need to parse this command

src/dotnet-uninstall/Shared/Exceptions/ExceptionHandler.cs
```diff
@@ -16,6 +19,17 @@ public static Action HandleException(Action action)
                 {
                     PrintExceptionMessage(e.Message);
                 }
+                catch (Win32Exception e)
```
wli3: This catch is too far, you should catch it in the immediate caller method(or the adapter layer). And handle it there. Following [this](https://docs.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions)

Currently it means the whole stack has the possibly to throw NATIVE_ERROR_CODE_CANCELED. It is hard to refactor. In higher level, you let the adapter error handling leaked to other place.

yuchong-pan: I agree. Will move this to a more proper place.

src/dotnet-uninstall/Windows/ProcessHandler.cs
```diff
@@ -0,0 +1,123 @@
...
+        private static void RunAsAdmin()
+        {
+            var location = Assembly.GetEntryAssembly().Location;
+
+            var process = new Process
+            {
+                StartInfo = new ProcessStartInfo
+                {
+                    FileName = "cmd.exe",
+                    Arguments = $"/c dotnet {location} {string.Join(" ", Environment.GetCommandLineArgs().Skip(1))}",
```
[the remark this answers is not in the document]
wli3: directly call dotnet is not very reliable when it is not on the path.

the following way is better
https://github.com/dotnet/cli/commit/496e4d97a021479b5f6179b17dbdfd6a6a2ce9c1#diff-05fe0ebb72ebcf3193f2fe991583b96fR39

src/dotnet-uninstall/Windows/ProcessHandler.cs
yuchong-pan: I agree. Probably I will just try to rename this method to reflect its purpose.

src/dotnet-uninstall/Windows/ProcessHandler.cs
yuchong-pan: `argc` is a commonly used variable for argument count, originated in C.

src/dotnet-uninstall/Windows/ProcessHandler.cs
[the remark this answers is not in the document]
yuchong-pan: Sorry, I don't quite understand this comment.

src/dotnet-uninstall/Shared/Exceptions/ExceptionHandler.cs
yuchong-pan: Done.

src/dotnet-uninstall/Windows/ProcessHandler.cs
yuchong-pan: I have renamed it `ParseCommandToArgs`, which I think is clear to convey what's going on.

src/dotnet-uninstall/Windows/ProcessHandler.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using System.Security.Principal;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Exceptions;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Windows
+{
+    public static class ProcessHandler
```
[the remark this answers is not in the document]
yuchong-pan: I agree. And it does things more than processes. I think I could rename this class to `CommandLineExec`.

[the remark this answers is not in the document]
yuchong-pan: Done.

src/dotnet-uninstall/Windows/ProcessHandler.cs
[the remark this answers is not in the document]
yuchong-pan: Replaced `dotnet` with what you suggested.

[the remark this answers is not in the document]
yuchong-pan: Hi @wli3, what do you mean by "directly execute windows installer uninstall command in admin without this tool get involved"?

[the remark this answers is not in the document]
wli3: replied in email

src/dotnet-uninstall/Windows/ProcessHandler.cs
wli3: not a fan of c, c++ naming style. but i can find it on the internet. so that is fine

PR #13, opened 2019-06-20
Architecture selection options

yuchong-pan:
Added architecture selection options (`--arm32`, `--x86`, `--x64`) for both of the `list` command and the `uninstall` command.

UI issues:
1. For the option combinations e.g. `--all-but-latest --x64 --sdk`, is the "latest" here mean the latest over all SDKs, or the latest over x64 SDKs only?
_Response from @KathleenDollard: The latest over x64 SDKs only._

_Note: This pull request is currently a draft and should be reviewed after Pull Request #12 has been merged into `master`._


src/dotnet-uninstall/Messages.cs
```diff
@@ -34,8 +34,11 @@ internal static class Messages
         public static readonly string UninstallVerbosityOptionArgumentName = "VERBOSITY-LEVEL";
         public static readonly string UninstallVerbosityOptionArgumentDescription = "The specified MSBuild verbosity level";
 
-        public static readonly string SdkOptionDescription = "Uninstall the .NET Core SDKs only";
-        public static readonly string RuntimeOptionDescription = "Uninstall the .NET Core Runtimes only";
+        public static readonly string SdkOptionDescription = "Uninstall .NET Core SDKs only";
```
wli3: you should start to convert to resx now. The longer you delay that, the harder it gets

src/dotnet-uninstall/Shared/BundleInfo/BundleVersion.cs
```diff
@@ -44,7 +44,7 @@ public static BundleVersion FromInput<TBundleVersion>(string input)
                 return RuntimeVersion.FromInput(input);
             }
 
-            throw new ArgumentOutOfRangeException();
+            throw new ArgumentException();
```
wli3: nit: even only the developer should see this exception, it is usually good to add some explaining why. So people can debug faster. You can add something like "cannot find enum matching"

test/dotnet-uninstall.Tests/Shared/Filterers/AllBelowOptionFiltererTests.cs
```diff
@@ -115,15 +122,114 @@ public static IEnumerable<object[]> GetDataForTestFiltererGood()
...
+                },
+                BundleType.Runtime,
+                BundleArch.X64 | BundleArch.Arm32
             };
         }
 
         [Theory]
         [MemberData(nameof(GetDataForTestFiltererGood))]
-        internal void TestAllBelowOptionFiltererGood(IEnumerable<Bundle> testBundles, string argValue, IEnumerable<Bundle> expected, BundleType typeSelection)
+        internal void TestAllBelowOptionFiltererGood(IEnumerable<Bundle> testBundles, string argValue, IEnumerable<Bundle> expected, BundleType typeSelection, BundleArch archSelection)
```
wli3: put expected to the end, everything before that is input. VS has automatic refactoring to change the signature, should be easy to do

src/dotnet-uninstall/Messages.cs
yuchong-pan: I agree. But my concern is that there are always pending PRs. I will probably do that during the weekend if we get all merged in today.

test/dotnet-uninstall.Tests/Shared/Filterers/AllBelowOptionFiltererTests.cs
yuchong-pan: Discussed offline. This will be too time-consuming for now.

PR #20, opened 2019-06-22
Support `<major>.<minor>` for the `--all-below` option

yuchong-pan:
@KathleenDollard mentioned during an offline conversation that it would be convenient for users to be able to specify `<major>.<minor>` only for the `--all-below` option. Hence I add the functionality.

The main approach to achieve this is to parse `<major>.<minor>` as `<major>.<minor>.0.0`, where the patch number and the build number are both `0`.

Covered by unit tests.

### DIFF
--- a/test/dotnet-uninstall.Tests/Shared/BundleInfo/SdkVersionTests.cs
+++ b/test/dotnet-uninstall.Tests/Shared/BundleInfo/SdkVersionTests.cs
@@ -199,6 +199,14 @@ namespace Microsoft.DotNet.Tools.Uninstall.Tests.Shared.BundleInfo
         [InlineData("2.2.300", 2, 2, 3, 0, 0, false)]
         [InlineData("3.0.100-preview-009812", 3, 0, 1, 0, 9812, true)]
         [InlineData("3.0.100-preview6-012264", 3, 0, 1, 0, 12264, true)]
+        [InlineData("1.0", 1, 0, 0, 0, 0, false)]
+        [InlineData("1.1", 1, 1, 0, 0, 0, false)]
+        [InlineData("2.0", 2, 0, 0, 0, 0, false)]
+        [InlineData("2.1", 2, 1, 0, 0, 0, false)]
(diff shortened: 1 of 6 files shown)


test/dotnet-uninstall.Tests/Shared/BundleInfo/SdkVersionTests.cs
```diff
@@ -199,6 +199,14 @@ internal void TestInequalityNull(SdkVersion version)
...
         [InlineData("3.0.100-preview6-012264", 3, 0, 1, 0, 12264, true)]
+        [InlineData("1.0", 1, 0, 0, 0, 0, false)]
+        [InlineData("1.1", 1, 1, 0, 0, 0, false)]
+        [InlineData("2.0", 2, 0, 0, 0, 0, false)]
+        [InlineData("2.1", 2, 1, 0, 0, 0, false)]
+        [InlineData("2.2", 2, 2, 0, 0, 0, false)]
+        [InlineData("3.0", 3, 0, 0, 0, 0, false)]
+        [InlineData("12.345", 12, 345, 0, 0, 0, false)]
+        [InlineData("0012.00345", 12, 345, 0, 0, 0, false)]
         internal void TestFromInputAccept(string input, int major, int minor, int sdkMinor, int patch, int build, bool preview)
```
wli3: nit: maybe https://www.nuget.org/packages/NuGet.Versioning/ can help you parse most of the stuff once you use regex to pick out the version. So you don't need to separate major and minor using regex

yuchong-pan: Thank you @wli3 for suggesting the NuGet package. However according to the description of the package, it seems that the package uses SemVer. Unfortunately SDK versions are not strictly SemVer (as discussed before, SDK versions have an extra "SDK minor"). Therefore, we can't use this to parse SDK versions.

If we use this package for Runtime versions and manual parsing for SDK versions, there are inconsistencies between two counterparts of the same thing. I'm wondering if this is good.

We could however definitely discuss this more, as .NET Core versioning is very different.

PR #21, opened 2019-06-24
Disallowed `--sdk` and `--runtime` to occur simultaneously

yuchong-pan:
Having had several offline conversations with @KathleenDollard, we eventually decide not to allow the `--sdk` option and the `--runtime` option to occur simultaneously, even for the options that do not have ambiguities between SDK versions and Runtime versions, such as `--all` and `--major-minor`.

The reason for this decision is that it will otherwise be very difficult to deliver the information about which options require exactly one of `--sdk` and `--runtime` and which don't. This will be extremely hard to the help information.

### DIFF
--- a/test/dotnet-uninstall.Tests/Shared/Filterers/MajorMinorOptionFiltererTests.cs
+++ b/test/dotnet-uninstall.Tests/Shared/Filterers/MajorMinorOptionFiltererTests.cs
@@ -10,7 +10,7 @@ namespace Microsoft.DotNet.Tools.Uninstall.Tests.Shared.Filterers
     public class MajorMinorOptionFiltererTests : FiltererTests
     {
         internal override Option Option => CommandLineConfigs.UninstallMajorMinorOption;
-        internal override string DefaultArgValue => "2.2";
+        internal override string DefaultTestArgValue => "2.2";
 
         public static IEnumerable<object[]> GetDataForTestFiltererGood()
         {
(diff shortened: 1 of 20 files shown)


wli3:
@KathleenDollard question about this UI design. If we use subcommand instead of option will it be more clear for the user? Since subcommand is exclusive by default. It might be more clear for both the user and to implement

test/dotnet-uninstall.Tests/Shared/Filterers/MajorMinorOptionFiltererTests.cs
```diff
@@ -208,7 +127,8 @@ internal void TestMajorMinorOptionFiltererGood(IEnumerable<Bundle> testBundles,
         [InlineData("3.0.100-preview5-011568")]
         internal void TestMajorMinorOptionFiltererInvalidInputVersionStringException(string argValue)
         {
-            TestFiltererException<InvalidInputVersionException>(DefaultTestBundles, argValue, BundleType.Sdk | BundleType.Runtime, DefaultTestArchSelection);
```
wli3: Is there a test for "when options have both --sdk and --runtime, it should throw"? If so just point me to it and it is good  to merge. If not, just add it and merge

yuchong-pan: Yes, it's in the base test class and applies to all filterer tests.

PR #23, opened 2019-06-25
Error message in place of self-elevation

yuchong-pan:


src/dotnet-uninstall/LocalizableStrings.Designer.cs
```diff
@@ -87,15 +87,6 @@ internal class LocalizableStrings {
             }
         }
         
-        /// <summary>
-        ///   Looks up a localized string similar to Failed to run as administrator.
-        /// </summary>
-        internal static string ElevationFailedExceptionMessage {
```
wli3: nit: delete this file. This is a bug in Arcade working with VS.

yuchong-pan: The solution does not build after this file is removed.

wli3: ok, so maybe we do need it

PR #27, opened 2019-06-27
WIP `FileSystemExplorer` for `list` command on macOS

yuchong-pan:
The reason for which `BundleVersion.SemVer` is no longer `readonly` is that C# allows generic type parameters to call `public` zero-argument constructors only.

### DIFF
--- a/src/dotnet-uninstall/MacOs/FileSystemExplorer.cs
+++ b/src/dotnet-uninstall/MacOs/FileSystemExplorer.cs
@@ -1,0 +1,42 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo.Versioning;
+
+namespace Microsoft.DotNet.Tools.Uninstall.MacOs
+{
+    internal static class FileSystemExplorer
+    {
+        private static readonly string DotNetInstallPath = Path.Combine("/", "usr", "local", "share", "dotnet");
+        private static readonly string DotNetSdkInstallPath = Path.Combine(DotNetInstallPath, "sdk");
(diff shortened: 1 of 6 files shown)


src/dotnet-uninstall/MacOs/FileSystemExplorer.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo.Versioning;
+
+namespace Microsoft.DotNet.Tools.Uninstall.MacOs
+{
+    internal static class FileSystemExplorer
```
wli3: https://github.com/dotnet/cli/blob/master/scripts/obtain/uninstall/dotnet-uninstall-pkgs.sh#L22

mac has pkgutil that is similar to Windows uninstaller. We should start from that first instead of directly look into the directory, there might be something left in pkgutil .

wli3: sorry i think i did not make it clear it earlier

src/dotnet-uninstall/Shared/BundleInfo/Versioning/BundleVersion.cs
```diff
@@ -51,6 +51,27 @@ public static BundleVersion FromInput<TBundleVersion>(string value)
             throw new ArgumentException();
         }
 
+        public static bool TryFromInput<TBundleVersion>(string value, out TBundleVersion version)
```
wli3: Is that "C# allows generic type parameters to call public zero-argument constructors only"? could you give me an example?

I don't think it matters much in this case, but in general, we should not expose an internal implementation as a public API due to a language constrain. Since we barely remove any public API after release, so the cost is very high

yuchong-pan:
Closed because of discussion on the macOS implementation.

PR #28, opened 2019-06-27
Refactoring of `BundleVersion.FromInput` & constructors of `SdkVersion` and `RuntimeVersion`

yuchong-pan:
From my point of view, this pull request provides a better workaround to address the issue brought up [here](https://github.com/dotnet/cli-lab/pull/22#discussion_r297475584).

_Note: This is currently a draft and should not be reviewed before Pull Request #27 has been merged into `master`._

PR #31, opened 2019-06-27
Refactoring of `BundleVersion.FromInput` & constructors of `SdkVersion` and `RuntimeVersion`

yuchong-pan:
From my point of view, this pull request provides a better workaround to address the issue brought up [here](https://github.com/dotnet/cli-lab/pull/22#discussion_r297475584).

I split Pull Request #28 to make it independent of macOS implementations. Hence it can be reviewed separately. Pull Request #28 has been closed.


yuchong-pan:
> although yesterday I have a comment that we should not make API public due to language constrains, could you reply that?

[Discussed offline.]

PR #34, opened 2019-07-02
`list` command for macOS

yuchong-pan:
Implemented the `list` command for macOS:
1. Added `ListCommandExec` for Windows and for macOS to abstract code that is common to both platforms.
2. Added `FileSystemExplorer` to access the bundle information installed on macOS.
3. Added `BundleTypePrintInfo` for future support of ASP.NET bundles.

### DIFF
--- a/src/dotnet-uninstall/MacOs/FileSystemExplorer.cs
+++ b/src/dotnet-uninstall/MacOs/FileSystemExplorer.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo.Versioning;
+
+namespace Microsoft.DotNet.Tools.Uninstall.MacOs
+{
+    internal static class FileSystemExplorer
+    {
+        private static readonly string DotNetInstallPath = Path.Combine("/", "usr", "local", "share", "dotnet");
+        private static readonly string DotNetSdkInstallPath = Path.Combine(DotNetInstallPath, "sdk");
+        private static readonly string DotNetRuntimeInstallPath = Path.Combine(DotNetInstallPath, "shared", "Microsoft.NETCore.App");
+        private static readonly string DotNetAspAllInstallPath = Path.Combine(DotNetInstallPath, "shared", "Microsoft.AspNetCore.All");
+        private static readonly string DotNetAspAppInstallPath = Path.Combine(DotNetInstallPath, "shared", "Microsoft.AspNetCore.App");
+        private static readonly string DotNetHostFxrInstallPath = Path.Combine(DotNetInstallPath, "host", "fxr");
+
+        public static IEnumerable<Bundle> GetInstalledBundles()
+        {
+            var sdks = GetInstalledBundles<SdkVersion>(DotNetSdkInstallPath);
+            var runtimes = GetInstalledBundles<RuntimeVersion>(
+                DotNetRuntimeInstallPath,
+                DotNetAspAllInstallPath,
+                DotNetAspAppInstallPath,
+                DotNetHostFxrInstallPath);
+
+            return sdks.Concat(runtimes);
+        }
+
+        private static IEnumerable<Bundle> GetInstalledBundles<TBundleVersion>(params string[] paths)
+            where TBundleVersion : BundleVersion, IComparable<TBundleVersion>, new()
+        {
+            return paths
+                .SelectMany(path => GetInstalledVersionsAndUninstallCommands<TBundleVersion>(path))
+                .GroupBy(tuple => tuple.Version)
+                .Select(group => Bundle.From(
+                    group.First().Version,
+                    BundleArch.X64,
+                    CombineUninstallCommands(group.Select(tuple => tuple.UninstallCommand)),
+                    ""));
+        }
+
+        private static IEnumerable<(TBundleVersion Version, string UninstallCommand)> GetInstalledVersionsAndUninstallCommands<TBundleVersion>(string path)
+            where TBundleVersion : BundleVersion, IComparable<TBundleVersion>, new()
+        {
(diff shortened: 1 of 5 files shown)


src/dotnet-uninstall/MacOs/FileSystemExplorer.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo.Versioning;
+
+namespace Microsoft.DotNet.Tools.Uninstall.MacOs
+{
+    internal static class FileSystemExplorer
+    {
+        private static readonly string DotNetInstallPath = Path.Combine("/", "usr", "local", "share", "dotnet");
+        private static readonly string DotNetSdkInstallPath = Path.Combine(DotNetInstallPath, "sdk");
+        private static readonly string DotNetRuntimeInstallPath = Path.Combine(DotNetInstallPath, "shared", "Microsoft.NETCore.App");
```
wli3: Could you tried to install 2.0, 2.1X, 2.2X and 3.0 and see what is the actual folder name. And I think we need to aggregate them in to one runtime.

wli3: I think Lee was talking about for mac, combine aspnet with runtime to make it just "runtime". I don't it is there

yuchong-pan: Yes, we combine ASP.NET Runtimes and Host FXRs with .NET Runtimes for macOS. This is the intended behavior. However, we still need to know which ASP.NET Runtimes and Host FXRs are installed in order to aggregate them. That's why we need the install paths for them here. If you run the tool with `dotnet uninstall list`, you will see that there are only two sections, namely SDKs and Runtimes.

src/dotnet-uninstall/Windows/ListCommandExec.cs
```diff
@@ -0,0 +1,37 @@
...
+            var gridView = new GridView();
+
+            gridView.SetColumns(Enumerable.Repeat(ColumnDefinition.SizeToContent(), 3).ToArray());
+            gridView.SetRows(Enumerable.Repeat(RowDefinition.SizeToContent(), Math.Max(bundles.Count, 1)).ToArray());
+
+            foreach (var (bundle, index) in bundles.Select((bundle, index) => (bundle, index)))
+            {
+                gridView.SetChild(new ContentView(string.Empty), 0, index);
+                gridView.SetChild(new ContentView(bundle.Version.ToString()), 1, index);
+                gridView.SetChild(new ContentView($"\"{bundle.DisplayName}\""), 2, index);
```
peterhuene: I know this is what we had before, but is there a reason to quote the display name in the output?

src/dotnet-uninstall/MacOs/ListCommandExec.cs
```diff
@@ -0,0 +1,37 @@
...
+            var gridView = new GridView();
+
+            gridView.SetColumns(Enumerable.Repeat(ColumnDefinition.SizeToContent(), 3).ToArray());
+            gridView.SetRows(Enumerable.Repeat(RowDefinition.SizeToContent(), Math.Max(bundles.Count, 1)).ToArray());
+
+            foreach (var (bundle, index) in bundles.Select((bundle, index) => (bundle, index)))
+            {
+                gridView.SetChild(new ContentView(string.Empty), 0, index);
+                gridView.SetChild(new ContentView(bundle.Version.ToString()), 1, index);
+                gridView.SetChild(new ContentView($"({bundle.Arch.ToString().ToLower()})"), 2, index);
```
peterhuene: For macOS, this will always output `x64`, right?  Is that useful information?  It appears that this line is the only different between this implementation and the one for Windows.

If we want to keep this, perhaps we should set the display name for the bundle to `BundleArch.X64.ToString().ToLowerInvariant()` (or simply just "x64") when creating them in `FileSystemExplorer.GetInstalledBundles` and thus could potentially unify the `ListCommandExec` implementations (i.e. always output `bundle.DisplayName`) unless we foresee different output based on platform for the future?

src/dotnet-uninstall/Windows/ListCommandExec.cs
yuchong-pan: I asked @KathleenDollard whether we should quote the display name, because otherwise the whitespace in the display name may be confusing together with column separators. I remember Kathleen thinks it is good to have double quotes around the display name.

peterhuene: Its a little weird to me in a grid view since this text is in a column, but I'm not hung up on this.

src/dotnet-uninstall/MacOs/ListCommandExec.cs
yuchong-pan: Yes, you're right - for macOS this will always output `x64`. However, I personally think it will still be useful information because we have `--x64` and `--x86` options. Users may get confused when looking at the help page if we don't have `x64` in the list output.

I'm not sure if putting `x64` in the display name is a good idea because `x64` is an architecture type but not a display name. We don't have a display name for macOS installations. If we put `x64` in `DisplayName`, it will be confusing for other developers of this tool. We could instead "fake" a display name such as "Microsoft .NET Core 2.2.300 (x64)` in place of the architecture type.

peterhuene: A derived display name might give a more consistent UX between platforms. @KathleenDollard what's your take?

yuchong-pan: I have indeed asked @KathleenDollard about this, and the answer was it's okay to have UI different between platforms. That being said, we could definitely tweak this more for a better UX.

peterhuene: I don't find "x64" repeated ad infinitum to be useful information, but if this has already been discussed with Kathleen, I won't hold up this PR.

PR #37, opened 2019-07-03
Verbosity levels

yuchong-pan:
Added `VerbosityLevel` and `VerbosityHandler` to execute different functions based on the specified verbosity level.


src/dotnet-uninstall/Shared/Configs/Verbosity/VerbosityHandler.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Shared.Configs.Verbosity
+{
+    internal class VerbosityHandler<TArg>
```
peterhuene: I think this is an interesting approach to what is essentially a logging API; however, I personally find this API a little hard to reason with.

* The API appears designed to register a single action per-verbosity level.  Would there be a case where I would want multiple actions at different verbosity levels to be executed at the same time via `Execute` (i.e. one action for normal and another action for minimal, but both should execute together when the configured verbosity level is minimal)?
* Given the way `VerbosityHandler` is used in the `foreach` loop in `UninstallCommandExec.cs`, the message that we log seems distant to where `VerbosityHandler.Execute` is called.  It may be difficult to figure out what `Execute` is doing since one needs to figure out what was registered.  I would like to see more locality regarding the action being performed, which I think should be restricted to logging messages.
* I think most developers would be more familiar with a traditional `Logger.LogXXX(...)` API, where the (global) "logger" was configured with the verbosity level to use.  In the CLI we have `Reporter.Verbose.WriteLine(...)`, but there's only a "verbose" and "not verbose" level in the CLI, so I wouldn't say that's a good example.  MSBuild, for example, has `LogMessage`, `LogWarning`, and `LogError` in the `TaskBase` class for logging messages at the appropriate severity and (internal to the logger implementation) verbosity level.  Unfortunately, C# doesn't have compile-time macros to facilitate lazy evaluation of the the message expression one wants to log, so one either needs to wrap it with a "level check" (e.g. `if (Logger.IsNormalVerbosityEnabled) { Logger.LogNormal("hello"); }`) or pass an action to the method (e.g. `Logger.LogNormal(() => "hello")`).
* I feel that the `VerbosityHandler` is a little too generic in intention such that actions can do whatever they want rather than being about logging messages at a given verbosity level.
* As a user, I would expect the `Quiet` verbosity level to never log; with this implementation it is possible to override the default "quiet" handler to do something.  Thus, my expectation would be that "quiet" never invokes any action.

yuchong-pan: Discussed offline. Replaced `VerbosityHandler` with `VerbosityLogger`.

PR #39, opened 2019-07-03
Flags attribute for `BundleType` and `BundleArch`

yuchong-pan:


src/dotnet-uninstall/Shared/BundleInfo/BundleArch.cs
```diff
@@ -1,5 +1,8 @@
-﻿namespace Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo
+﻿using System;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo
 {
+    [Flags]
```
wli3: haven't use this myself. maybe talk to Nick for the best practice? But that's fine to checkin since we already have good test coverage

PR #40, opened 2019-07-08
aka.ms link update

yuchong-pan:
As discussed with @KathleenDollard and Rowan during an offline meeting, the name of the tool will be `dotnet-core-uninstall` and the aka.ms link will be "aka.ms/dotnet-core-uninstall".

I will create an issue to setup the aka.ms link as well.


yuchong-pan:
Issue created at #41.

PR #48, opened 2019-07-10
Dry run do-it command fix

yuchong-pan:
Before:
```
*** DRY RUN
Would uninstall: .NET Core SDK 1.1.14 (x64).
*** END DRY RUN

To actually uninstall, add --do-it: C:\Users\t-yup\github\cli-lab\artifacts\bin\dotnet-core-uninstall\Release\netcoreapp2.1\win-x64\publish\dotnet-core-uninstall.dll--major-minor 1.1 --sdk --do-it.
```

After:
```
*** DRY RUN
Would uninstall: .NET Core SDK 1.1.14 (x64).
*** END DRY RUN

To actually uninstall, add --do-it: C:\Users\t-yup\github\cli-lab\artifacts\bin\dotnet-core-uninstall\Release\netcoreapp2.1\win-x64\publish\dotnet-core-uninstall.exe --major-minor 1.1 --sdk --do-it.
```

### DIFF
--- a/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
+++ b/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
@@ -7,6 +7,8 @@ using Microsoft.DotNet.Tools.Uninstall.Shared.Utils;
 using Microsoft.DotNet.Tools.Uninstall.Windows;
 using System.Reflection;
 using Microsoft.DotNet.Tools.Uninstall.MacOs;
+using System.Linq;
+using System.Diagnostics;
 
 namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
 {
@@ -108,7 +110,7 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
             Console.WriteLine();
             Console.WriteLine(string.Format(
                 LocalizableStrings.DryRunHowToDoItMessageFormat,
-                string.Join(" ", Environment.GetCommandLineArgs())));
+                $"{Process.GetCurrentProcess().MainModule.FileName} {string.Join(" ", Environment.GetCommandLineArgs().Skip(1))}"));
         }
 
         private static void HandleVersionOption()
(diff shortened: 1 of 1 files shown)


src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
```diff
@@ -108,7 +110,7 @@ private static void TryIt(IEnumerable<Bundle> bundles)
             Console.WriteLine();
             Console.WriteLine(string.Format(
                 LocalizableStrings.DryRunHowToDoItMessageFormat,
-                string.Join(" ", Environment.GetCommandLineArgs())));
+                $"{Process.GetCurrentProcess().MainModule.FileName} {string.Join(" ", Environment.GetCommandLineArgs().Skip(1))}"));
```
peterhuene: I'd suggest this as the full path isn't usually necessary when invoking command line tools:

```suggestion
                $"{Path.GetFileName(Process.GetCurrentProcess().MainModule.FileName)} {string.Join(" ", Environment.GetCommandLineArgs().Skip(1))}"));
```

yuchong-pan: However, I think this would be confusing if the user is _not_ running the executable in its directory.

yuchong-pan: I'm not sure if we are going to ship a single executable or an MS installer to install it to some `PATH` location.

peterhuene: I feel the output to be too long and not generally helpful.  I would not expect users to "copy and paste" the output to re-run with the `--do-it` option; they'd do what most developers would do: "hit up and type `--do-it`".

However, I don't have a strong opinion.

peterhuene: Likewise, I would also be fine with a simple (and static) message of `To actually perform the uninstall, use the '--do-it' option.`  The command's sitting in my shell history, and I know where to put the option :)

yuchong-pan: Done.

PR #49, opened 2019-07-12
Support for ASP.NET Runtimes

yuchong-pan:
Added support for ASP.NET Runtimes, with the `--aspnet-runtime` option.


src/dotnet-core-uninstall/Shared/Utils/Regexes.cs
```diff
@@ -39,26 +47,34 @@ internal static class Regexes
...
+            _runtimeVersionBasicRegexFormat,
+            _previewVersionAspNetRuntimeDisplayNameRegex.ToString()));
         private static readonly Regex _sdkDisplayNameRegex = new Regex(
             $@"\.NET\sCore\s(?<{TypeGroupName}>SDK)\s{_sdkVersionDisplayNameRegex.ToString()}|Microsoft\s\.NET\sCore\s(?<{TypeGroupName}>SDK)\s(\-\s)?{_sdkVersionDisplayNameRegex.ToString()}");
         private static readonly Regex _runtimeDisplayNameRegex = new Regex(
             $@"Microsoft\s\.NET\sCore\s(?<{TypeGroupName}>Runtime)\s\-\s{_runtimeVersionDisplayNameRegex.ToString()}|Microsoft\s\.NET\sCore\s{_runtimeVersionDisplayNameRegex.ToString()}\s\-\s(?<{TypeGroupName}>Runtime)");
+        private static readonly Regex _aspNetRuntimeDisplayNameRegex = new Regex(
+            $@"Microsoft\s(?<{TypeGroupName}>ASP.NET)\sCore\s{_aspNetRuntimeVersionDisplayNameRegex.ToString()}\s\-\s(Shared\sFramework|Runtime\sPackage\sStore)");
         private static readonly Regex _archRegex = new Regex(
-            $@"(?<{ArchGroupName}>(x64|x86|arm32))");
```
peterhuene: Double checking that the removal of `arm32` was intentional.  I assume the tool is just going to focus on x86/x64 for now?

peterhuene:
LGTM 👍

src/dotnet-core-uninstall/Shared/Utils/Regexes.cs
yuchong-pan: Yes. Actually we do not support `arm32` because there are no arm32 .NET Core bundles. I removed other arm32-related stuff in previous PRs, but I just realized I forgot to remove it from Regexes.

PR #70, opened 2019-08-02
Construction of target graph

yuchong-pan:
_Note: Unit tests are not complete._
~~_Note: This PR should be reviewed after #69 is merged._~~

### DIFF
--- a/src/MSBuildBinLogQuery/Construction/GraphBuilder.cs
+++ b/src/MSBuildBinLogQuery/Construction/GraphBuilder.cs
@@ -43,11 +43,11 @@ namespace Microsoft.Build.Logging.Query.Construction
         private void TargetStarted(object sender, TargetStartedEventArgs args)
         {
             var project = Projects[args.BuildEventContext.ProjectInstanceId];
-            var target = project.AddOrGetTarget(args.TargetName);
+            var target = project.AddOrGetOrderedTarget(args.TargetName);
 
             if (!string.IsNullOrWhiteSpace(args.ParentTarget))
             {
-                var parent = project.AddOrGetTarget(args.ParentTarget);
+                var parent = project.AddOrGetOrderedTarget(args.ParentTarget);
 
                 if (args.BuildReason == TargetBuiltReason.DependsOn)
                 {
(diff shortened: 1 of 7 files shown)


src/MSBuildBinLogQuery/Graph/ProjectNode.cs
```diff
@@ -13,9 +16,9 @@ public class ProjectNode : IQueryableGraphNode
         public ItemManager Items { get; }
         public PropertyManager Properties { get; }
         public PropertyManager GlobalProperties { get; }
-        public List<TargetNode> ExecutedTargets { get; }
-        public List<TargetNode> EntryPointTargets { get; }
-        public HashSet<ProjectNode> ProjectsBeforeThis { get; }
+        public ConcurrentDictionary<string, TargetNode> Targets { get; }
+        public ImmutableHashSet<TargetNode> EntryPointTargets { get; }
```
rainersigwald: I think this should be an `ImmutableList`, even though managing it will be harder. The order can be relevant (in some cases).

_Really_ it should be an unordered set of ordered lists, but let's hope we never have to be that accurate.

src/MSBuildBinLogQuery/Construction/GraphBuilder.cs
```diff
@@ -32,13 +30,38 @@ public void HandleEvents(params BuildEventArgs[] buildEvents)
         private void ProjectStarted(object sender, ProjectStartedEventArgs args)
         {
             var id = args.BuildEventContext.ProjectInstanceId;
-            Projects.GetOrAdd(id, new ProjectNode(id, args));
+            Projects.TryAdd(id, new ProjectNode(id, args));
```
rainersigwald: This is interesting. It's ok for now, but please add a TODO comment:

> A single project instance can be built concurrently with distinct EntryPointTargets. Those should be tracked as part of the overall ProjectNode.

src/MSBuildBinLogQuery/Graph/ProjectNode.cs
yuchong-pan: Done. Actually I made it to an `IReadOnlyList`.

src/MSBuildBinLogQuery/Construction/GraphBuilder.cs
yuchong-pan: Done.

PR #72, opened 2019-08-06
DAG dependencies

yuchong-pan:
_Note: This PR should be reviewed after #70 has been merged into `master`._

1. class for directed acyclic graphs (DAG)
2. obtaining the topological ordering of a DAG
3. obtaining the reverse graph of a DAG
4. obtaining reachable nodes of each node in a DAG (for calculating _all_ dependencies of a project/target).

PR #75, opened 2019-08-09
Messaging handling

yuchong-pan:
~~_Note: This PR should be reviewed after #72 has been merged into `master`._~~

1. event handling of tasks, messages, warnings and errors
2. All messages, warnings, tasks under a component (`BuildComponent.AllMessages`, `BuildComponent.AllWarnings`, `BuildComponent.AllErrors`)

### DIFF
--- a/src/MSBuildBinLogQuery/Component/Component.cs
+++ b/src/MSBuildBinLogQuery/Component/Component.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using Microsoft.Build.Logging.Query.Messaging;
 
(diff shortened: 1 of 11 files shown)


src/MSBuildBinLogQuery/Component/Component.cs
```diff
@@ -0,0 +1,67 @@
...
+        public void AddWarning(Warning warning)
+        {
+            _warnings.Add(warning);
+            GoUpToRoot(component => component._allWarnings.Add(warning));
+        }
+
+        public void AddError(Error error)
+        {
+            _errors.Add(error);
+            GoUpToRoot(component => component._allErrors.Add(error));
```
rainersigwald: This is fine for now but in the future it may be better to compute this top-down via child relationships instead of storing it bottom-up as here.

yuchong-pan: Do you mean computing `AllMessages` and other stuff by a depth-first search recursively? I considered it, but that will have a worse time complexity. The bottom-up approach is O(1), but the top-down would be O(n).

rainersigwald: True. I don't know how often we will want to do this operation specifically, so we can decide on the memory/time tradeoff later. I suspect that we'll wind up doing so many other graph-walking operations that this won't seem worth treating specially.

yuchong-pan: Got it. We can refactor this later.

PR #79, opened 2019-08-13
Scanner of MSBuild binlog query language

yuchong-pan:
Tokens supported (case insensitive):
* `MESSAGE`
* `WARNING`
* `ERROR`
* `PROJECT`
* `TARGET`
* `TASK`
* `/`
* `//`
* `[`
* `]`
* `=`
* `,`
* `"<STRING>"`


test/MSBuildBinLogQuery.CommandLine/Program.cs
```diff
@@ -29,6 +31,13 @@ public static void Main(string[] args)
             var build = graphBuilder.HandleEvents(events.ToArray());
 
             PrintProjectNodes(build);
+
+            var scanner = new Scanner(args[1]);
+
+            for (; !(scanner.Token is EofToken); scanner.ReadNextToken())
+            {
+                Console.WriteLine(scanner.Token + (scanner.Token is StringToken ? $" {(scanner.Token as StringToken).Value}" : ""));
```
rainersigwald: nit:

```suggestion
                Console.WriteLine(scanner.Token + (scanner.Token is StringToken ? $" {(scanner.Token as StringToken)?.Value ?? string.empty}"));
```

src/dotnet-core-uninstall/dotnet-core-uninstall.csproj
```diff
@@ -37,11 +37,7 @@
   </ItemGroup>
 
   <Target Name="PublishSingleFile" AfterTargets="Build">
-    <MSBuild
-      BuildInParallel="False"
-      Projects="$(MSBuildThisFileDirectory)\dotnet-core-uninstall.csproj"
-      Properties="NoBuild=true"
-      Targets="Publish">
+    <MSBuild BuildInParallel="False" Projects="$(MSBuildThisFileDirectory)\dotnet-core-uninstall.csproj" Properties="NoBuild=true" Targets="Publish">
```
rainersigwald: Why is this changing in this PR?

src/MSBuildBinLogQuery/Token/TaskToken.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        public bool Equals([AllowNull] TaskToken other)
+        {
+            return other != null;
+        }
+
+        public override int GetHashCode()
+        {
+            return 1;
```
rainersigwald: Is this going to cause problems if you have a `Dictionary<Token, _>`? I think everything would collide. Maybe instead switch to 
```suggestion
            return nameof(Token).GetHashCode();
```
? That would at least spread these out.

src/MSBuildBinLogQuery/Token/ErrorToken.cs
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            return Equals(obj as ErrorToken);
+        }
+
+        public bool Equals([AllowNull] ErrorToken other)
+        {
+            return other != null;
```
rainersigwald: Could this repetitive implementation be simplified by adding a new `ConstantToken` class that does the comparisons?

src/MSBuildBinLogQuery/Scan/Scanner.cs
```diff
@@ -0,0 +1,168 @@
...
+                return true;
+            }
+
+            _char = '\0';
+            return false;
+        }
+
+        private void SkipWhiteSpace()
+        {
+            while (IsWhiteSpace(_char) && ReadNextCharacter());
```
rainersigwald:
```suggestion
            while (IsWhiteSpace(_char) && ReadNextCharacter()) { }
```
This is more clear to me as a do-nothing loop.

src/MSBuildBinLogQuery/Token/TaskToken.cs
yuchong-pan: Should it be `nameof(TaskToken).GetHashCode()`?

src/dotnet-core-uninstall/dotnet-core-uninstall.csproj
yuchong-pan: I think this was just caused by the formatting of the editor. The content is exactly the same.

src/MSBuildBinLogQuery/Token/ErrorToken.cs
yuchong-pan: Do you mean an `enum` field or a string field inside `ConstantToken`? I still prefer to have `MessageToken`, `ErrorToken`, etc. separately because that provides more compile-time verifications.

src/MSBuildBinLogQuery/Token/TaskToken.cs
nguerrera: Haven't read all the code or understood the rationale, yet, but IEquatable + inheritance is perilous IMHO.

yuchong-pan: Done.

nguerrera: Oh, I see it is only on the leaf nodes. Consider sealing anything with value semantics.

yuchong-pan: Do you mean sealing all the leaf nodes here?

src/MSBuildBinLogQuery/Token/ErrorToken.cs
rainersigwald: No, I like separate classes, too. Just wondering if they could have a common base type to do the identical parts of their implementation.

src/dotnet-core-uninstall/dotnet-core-uninstall.csproj
rainersigwald: Please revert--the other way is more readable :)

src/MSBuildBinLogQuery/Scan/Scanner.cs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        public void ReadNextToken()
+        {
+            SkipWhiteSpace();
+
+            switch (_char)
+            {
+                case '\0':
+                    Token = new EofToken();
```
nguerrera: I would suggest making all of these that are effectively singletons by their value semantics actual singletons in memory. In other words, make the constructors private and have EofToken.Instance, etc.

Then you could restrict the IEquatable to the things that have more than 1 value like StringToken. For the single value ones, the inherited reference equality semantics would then be fine.

This will save on boilerplate in the token classes, and save on heap allocations during scanning.

rainersigwald: This is a great idea

src/MSBuildBinLogQuery/Token/ErrorToken.cs
yuchong-pan: Would it be hard because they need to implement `IEquatable` and hence implement `Equals` with different parameter types?

src/dotnet-core-uninstall/dotnet-core-uninstall.csproj
yuchong-pan: Done.

src/MSBuildBinLogQuery/Token/Token.cs
```diff
@@ -0,0 +1,6 @@
+namespace Microsoft.Build.Logging.Query.Token
+{
+    public abstract class Token
+    {
```
nguerrera: Consider internal constructor here to express that the set of tokens is not open ended. All derived classes are in this assembly and sealed. Really, this is a discriminated union expressed as an object-oriented hierarchy.

Depending on how you expect this language to evolve, you might also consider expressing the discriminated union as a single struct Type with a Kind enum and string Text value. It's less pretty, but would further cut down on heap allocations, and switching over the kind is going to be a lot faster than doing type checks.

src/MSBuildBinLogQuery/Token/ErrorToken.cs
nguerrera: I like the singleton option better as it's much more straight forward, but for the record, you could use CRTP to solve that part. https://stackoverflow.com/a/2032759

src/MSBuildBinLogQuery/Token/Token.cs
yuchong-pan: Yeah, I actually considered a `Kind` enum and a string `Text` value. But C# doesn't have a `Void` or `None` type which is awkward...

src/MSBuildBinLogQuery/Scan/Scanner.cs
yuchong-pan: Done.

src/MSBuildBinLogQuery/Token/WarningToken.cs
```diff
@@ -1,27 +1,24 @@
...
-        public WarningToken() : base()
+        public static WarningToken Instance
         {
-        }
+            get
+            {
+                if (_instance == null)
+                {
+                    _instance = new WarningToken();
+                }
```
nguerrera: I would suggest this:

``` C#
public static WarningToken Instance { get; } = new WarningToken();
```

The way you have it has a race condition if called on multiple threads.  This will be faster to access too with no branching. I don't think the laziness is necessary since the operation is so cheap. Above will allow the runtime to run the WarningToken when it sees fit as long as it is before the first use of static field. In practice, all of the constructors will run once during JIT compilation of the scanner method that uses Instance, and then the native code can just retrieve the static instance without a branch.

src/MSBuildBinLogQuery/Token/Token.cs
nguerrera: > But C# doesn't have a Void or None type which is awkward...

How is that related to the choice of a struct with Kind and Text vs. the class hierarchy. For the hierarchy, you are using `null` as your "none". You could do that with the struct too, if you used `Token?`. Or you could have a None kind.

yuchong-pan: I think what I want here is a class `Token<TValue>` with the type parameter `TValue`. So for the string token it would be `Token<string>` and for the integer token (to be added) it would be `Token<int>`. For others, ideally I want `Token<None>` or something similar (like `Void` in Java), but C# doesn't have this.

PR #80, opened 2019-08-13
Parser of MSBuild binlog query language 

yuchong-pan:
~~_Note: This PR should be reviewed after #79 has been merged._~~
~~_This PR is still a work in progress._~~

Grammer supported:
```
<SLASHES>      ::= "/"
                 | "//"
<MSG_NODE>     ::= <SLASHES> "message"
                 | <SLASHES> "warning"
                 | <SLASHES> "error"
<TASK_NODE>    ::= "/" "task"
                 | "/" "task" <MSG_NDOE>
<TARGET_NODE>  ::= "/" "target"
                 | "/" "target" <TASK_NODE>
                 | "/" "target" <MSG_NODE>
<PROJECT_NODE> ::= "/" "project"
                 | "/" "project" <TARGET_NODE>
                 | "/" "project" <TASK_NODE>
                 | "/" "project" <MSG_NODE>
<QUERY>        ::= <PROJECT_NODE>
                 | <TARGET_NODE>
                 | <TASK_NODE>
                 | <MSG_NODE>
```

Refactored LL(1) grammar:
```
<SLASHES>      ::= "/"
                 | "//"
<TEMP1>        ::= "message"
                 | "warning"
                 | "error"
<MSG_NODE>     ::= <SLASHES> <TEMP1>
<TEMP2>        ::= empty
                 | <MSG_NODE>
<TEMP4>        ::= <TEMP1>
                 | "task" <TEMP2>
<TEMP5>        ::= empty
                 | "/" <TEMP4>
                 | "//" <TEMP1>
<TEMP7>        ::= <TEMP1>
                 | "target" <TEMP5>
                 | "task" <TEMP2>
<TEMP8>        ::= empty
                 | "/" <TEMP7>
                 | "//" <TEMP1>
<TEMP9>        ::= "project" <TEMP8>
                 | "target" <TEMP5>
                 | "task" <TEMP2>
                 | <TEMP1>
<QUERY>        ::= "/" <TEMP9>
                 | "//" <TEMP1>
```

PR #82, opened 2019-08-16
Interpreter of MSBuild binlog query language

yuchong-pan:
_Note: This PR should be reviewed after #80 gets merged into `master`._

Practically, the only thing supported now is to distinguish messages/warnings/errors emitted by a project/target/task, respectively.

E.g. `//message` returns all messages that the build generates, while `/message` only returns top-level messages emitted by the build engine itself (such as the summary table) but not by any of the projects.


yuchong-pan:
Closing this PR because #84 has massive changes.

PR #93, opened 2019-09-19
Produce msi on Windows

wli3:
For dotnetCoreUninstall

Files under src/core-sdk-tasks is a straight copy from core-sdk.


wli3:
@johnbeisner could you review this PR?
Install/uninstall works, including add the exe to PATH.

Files under src/core-sdk-tasks is a copy from core-sdk without any change.

Also verified signing https://dnceng.visualstudio.com/internal/_build/results?buildId=360484

cc @livarcocc @yuchong-pan

src/redist/targets/packaging/windows/dotnetCoreUninstall/registrykeys.wxs
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0"?>
```
johnbeisner: There's nothing here - delete the file and reference.

eng/Versions.props
```diff
@@ -6,7 +6,10 @@
   </PropertyGroup>
   <!-- Repo Version Information -->
   <PropertyGroup>
-    <VersionPrefix>0.1.0</VersionPrefix>
+    <VersionMajor>0</VersionMajor>
```
johnbeisner: Is a "0" a legal major number?

src/redist/targets/GenerateMSIs.targets
```diff
@@ -0,0 +1,91 @@
...
+          DependsOnTargets="GenerateLayout;SetupWixProperties;GetDotnetCoreUninstallGitCommitInfo">
+
+    <GenerateMsiVersion CommitCount="$(GitCommitCount)"
+                        VersionMajor="$(VersionMajor)"
+                        VersionMinor="$(VersionMinor)"
+                        VersionPatch="$(VersionPatch)">
+      <Output TaskParameter="MsiVersion" PropertyName="MsiVersion" />
+    </GenerateMsiVersion>
+
+    <GenerateGuidFromName Name="$(DotnetCoreUninstallMSIInstallerFile)">
```
johnbeisner: What you a saying here with the upgradecode is that each MSI is a unique product - no upgrade in-place. That necessary means the installation paths need to be different or if not different, the version of the file(s) needs to be greater than the one before it.

src/redist/targets/packaging/windows/dotnetCoreUninstall/dotnet.wxs
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
+  <?include "Variables.wxi" ?>
+  <Product Id="*" Name="$(var.ProductName)" Language="$(var.ProductLanguage)" Version="$(var.ProductVersion)" Manufacturer="$(var.Manufacturer)" UpgradeCode="$(var.UpgradeCode)">
+    <Package Compressed="yes" InstallScope="perMachine" InstallerVersion="200" />
+
+    <MajorUpgrade DowngradeErrorMessage="$(var.DowngradeErrorMessage)" Schedule="afterInstallInitialize"/>
+
+    <?if $(var.Platform)=x86?>
```
johnbeisner: Are you building 2 installers - one for x86 and one for x64?

src/redist/targets/packaging/windows/dotnetCoreUninstall/dotnet.wxs
```diff
@@ -0,0 +1,50 @@
...
+    </Feature>
+    <Property Id="ProductFamily" Value="$(var.ProductFamily)" />
+    <Property Id="ProductEdition" Value="$(var.ProductEdition)" />
+    <Property Id="ProductCPU" Value="$(var.Platform)" />
+    <Property Id="RTM_ProductVersion" Value="$(var.Dotnet_ProductVersion)" />
+    <Property Id="MSIFASTINSTALL" Value="7" />
+    <WixVariable Id="WixUILicenseRtf" Value="$(var.MicrosoftEula)" />
+
+    <Property Id="DOTNETCOREUNINSTALLTEXE">
+      <DirectorySearch Id="DOTNETPATH" Path="[#DOTNETDOTNETCOREUNINSTALLHOME]">
```
johnbeisner: How are you setting the install directory if the 'InstallSharedHostandDetectionKeys' fails to find a host? Or do you plan to error the install and quit?

johnbeisner: Also why is it important to install where the ShardHost is? Can't we just install to ProgramFiles?

eng/Versions.props
wli3: yes. for example https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

src/redist/targets/GenerateMSIs.targets
wli3: hm... that's not what i want. It should just the bigger version can upgrade the older version. So I should just put the same guid here all the time?

src/redist/targets/packaging/windows/dotnetCoreUninstall/dotnet.wxs
wli3: no, but i don't want to remove 64 part. Since i could if I add a new leg

src/redist/targets/packaging/windows/dotnetCoreUninstall/dotnet.wxs
wli3: ah, i get it. I could just remove this part. I didn't really understand what these code was doing

src/redist/targets/GenerateMSIs.targets
johnbeisner: What is the installation behavior you want? I think this will work as long as the file versions keep going up - an example of this is the SharedHost MSI.

src/redist/targets/packaging/windows/dotnetCoreUninstall/registrykeys.wxs
wli3: removed

src/redist/targets/packaging/windows/dotnetCoreUninstall/dotnet.wxs
johnbeisner: Sounds good to me.

src/redist/targets/GenerateMSIs.targets
wli3: I just put a hard coded value. Is should be good?

src/redist/targets/packaging/windows/dotnetCoreUninstall/dotnet.wxs
wli3: Removed

johnbeisner: You should use:
ProgramFilesFolder - gets the well-known folder for CSIDL_PROGRAM_FILESX86.
ProgramFiles64Folder - gets the well-known folder for CSIDL_PROGRAM_FILES.
for the installation directory.

Something like this:
```
<?if $(var.Platform)=x86?>
DOTNETDOTNETCOREUNINSTALLHOME = ProgramFilesFolder
<?elseif $(var.Platform)=x64?>
DOTNETDOTNETCOREUNINSTALLHOME = ProgramFiles64Folder
<?endif?>
```

johnbeisner: @joeloff 
Can you review this PR?

PR #95, opened 2019-10-04
Vs version protection

sfoslund:
Protects Visual Studio from breaking by leaving the highest .NET Core SDK in each band (with special cases for 2.1.5xx and 2.2.1xx). Prevents uninstallation of these SDKs and removes them from list command unless option --include-required is included in command.


wli3:
@sfoslund forget to mention, usually we also add @dotnet/dotnet-cli  so everyone can get the notification

src/dotnet-core-uninstall/Shared/VSVersioning/VSVersionHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo.Versioning;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Exceptions;
+using NuGet.Versioning;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Shared.VSVersioning
+{
+    /// <summary>
+    /// Extracts VS versions and determine which sdk's are safe to uninstall
+    /// </summary>
+    internal static class VSVersionHelper
```
wli3: Try to reduce comments as possible. Use method name and code structure itself to convey the info of the comments until it is too complex to do so.

Usually BlahHelper contains no extra information. "Helper" part can just be removed. In this case, let's try to condense the comment to name. VisualStudioSafeVersionsExtracter ? It might be useful to point to Kathleen's post on why it is not safe. I'll find the post later

src/dotnet-core-uninstall/Shared/VSVersioning/VSVersionHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        /// <summary>
+        /// Divisions within version bands. Not inclusive, groupings: [2.1.000, 2.1.600) [2.1.600, 2.2.00) [2.0.00, 2.2.200) [2.2.200, 2.3.00)
+        /// </summary>
+        private static readonly SemanticVersion[] SpecialCaseDivisions = { new SemanticVersion(2, 1, 600), new SemanticVersion(2, 2, 200) };
```
wli3: This comment is a necessary one

src/dotnet-core-uninstall/Shared/VSVersioning/VSVersionHelper.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        /// <summary>
+        /// Assigns the uninstallAllowed flag to false for the highest sdkVersion in each band (Major.Minor) to ensure vs still functions properly
+        /// Note that 2.1.5/2.1.6 and 2.2.1/2.2.2 count as band cutoffs 
+        /// </summary>
+        public static void AssignUninstallAllowed(IEnumerable<Bundle> bundles)
```
wli3: "Assigns the uninstallAllowed flag to false for the highest sdkVersion in each band (Major.Minor) to ensure vs still functions properly" part is not necessary. It is almost a one to one translation to english of the linq expression

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtracter.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
+using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo.Versioning;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Exceptions;
+using NuGet.Versioning;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Shared.VSVersioning
+{
+    internal static class VisualStudioSafeVersionsExtracter
```
wli3: nit: Google told me it is Extract**o**r   :p

src/dotnet-core-uninstall/Shared/BundleInfo/Bundle.cs
```diff
@@ -11,6 +11,7 @@ internal abstract class Bundle : IEquatable<Bundle>
         public BundleArch Arch { get; }
         public string UninstallCommand { get; }
         public string DisplayName { get; }
+        public bool UninstallAllowed { get; set; }
```
wli3: Add this field in the class is not ideal. One code smell is it breaks encapsulation -- a big chuck of the operation logic on this field is in a different class VisualStudioSafeVersionsExtracter.

There are 2 way to go around it.

1. Remove this field. And simply run the VisualStudioSafeVersionsExtracter logic every time we need it. The downside is perf and duplication, we may need to compute twice. However, I think this is fine, considering this is a pure compute and we don't really have a time to compute it twice in one process live time. I think it fits this better, since the current code is very functional style.

2. Move VisualStudioSafeVersionsExtracter to Bundle. This is more of an "OOP" way of doing it. However, that would make Bundle class bigger. And currently non of the filtering operation is on Bundle class. It is odd for this code base.

test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            var bundles = new List<Bundle>();
+            foreach (string v in versions)
+            {
+                bundles.Add(new Bundle<SdkVersion>(new SdkVersion(v), new BundleArch(), string.Empty, v));
+            }
+
+            VisualStudioSafeVersionsExtracter.AssignUninstallAllowed(bundles);
+            // None of the bundles are uninstallable-> throw error
+            var exception = Assert.Throws<UninstallationNotAllowedException>(() => VisualStudioSafeVersionsExtracter.RemoveUninstallableBundles(bundles));
```
wli3: nit: Prefer using fluentassertions, here is an example of asserting exception https://github.com/dotnet/cli/blob/95c6eff6daa1a69f29c42b2d405400ad44bdec91/test/Microsoft.DotNet.ToolPackage.Tests/ToolConfigurationDeserializerTests.cs#L29

wli3:
Just finished one round of review. Very well done!

sfoslund:
@wli3 Thanks for your comments, your suggestions make things a lot cleaner. I also removed the --include-required option after discussing it with Kathleen yesterday and made displaying the required bundles the default.

src/dotnet-core-uninstall/MacOs/FileSystemExplorer.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection.Metadata.Ecma335;
```
wli3: Accidental check-in?

Issue #89, opened 2019-08-24
dotnet-core-uninstall might consider the INVERSE of --dry-run


shanselman:
PowerShell uses --whatif, by the way, as a standard.

We might consider /force or /doit. There's too much risk to "Accidentally" hurt myself.

KathleenDollard:
Let's change from options to verbs to indicate what will be done:

```
dotnet-core-uninstall list
dotnet-core-uninstall dry-run
dotnet-core-uninstall what
dotnet-core-uninstall remove
```

`dry-run` and `whatif` would be synonyms. 

I hate that one is posix and the other isn't, but if we include `whatif` it should be just like PowerShell

I believe this answers @shanselman's concern by avoiding a destructive default.

sfoslund:
Addressed in Restructure Commands PR (#96 )

PR #96, opened 2019-10-10
Restructure commands

sfoslund:
Restructure commands based on issue #89: Converted root command to "remove", converted option "--dry-run" to command, and added "whatif" command. 
Added upper limit for uninstallation: tool cannot be used for uninstalling sdk version 3.0.0 and above.
Allow explicit uninstallation of required bundles with extra warnings and confirmations.

### DIFF
--- a/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
+++ b/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
@@ -48,7 +48,7 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
                     throw new NotAdminException();
                 }
 
-                if (AskIt(filtered))
+                if (AskItAndReturnUserAnswer(filtered))
                 {
                     if (AskWithWarningsForRequiredBundles(filtered))
                     {
@@ -207,16 +207,20 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
             return args;
         }
 
-        private static bool AskIt(IDictionary<Bundle, string> bundles)
+        private static bool AskItAndReturnUserAnswer(IDictionary<Bundle, string> bundles)
         {
             var displayNames = string.Join("\n", bundles.Select(bundle => $"  {bundle.Key.DisplayName}"));
             Console.Write(string.Format(LocalizableStrings.ConfirmationPromptOutputFormat, displayNames));
 
             var response = Console.ReadLine().Trim().ToUpper();
 
-            if (response.Equals("Y") || response.Equals("N"))
+            if (response.Equals("Y"))
             {
-                return response.Equals("Y");
+                return true;
+            }
+            else if (response.Equals("N"))
+            {
+                return false;
             }
             else
             {
--- a/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
+++ b/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
@@ -30,7 +30,7 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
         {
             CommandBundleFilter.HandleVersionOption();
 
-            var filtered = CommandBundleFilter.GetFilteredBundles();
+            var filtered = CommandBundleFilter.GetFilteredWithRequirementStrings();
 
             if (CommandLineConfigs.CommandLineParseResult.CommandResult.OptionResult(CommandLineConfigs.YesOption.Name) != null)
             {
@@ -39,7 +39,7 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
                     throw new NotAdminException();
                 }
 
-                DoIt(filtered);
+                DoIt(filtered.Keys);
             }
             else
             {
@@ -48,7 +48,13 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
                     throw new NotAdminException();
                 }
 
-                AskIt(filtered);
+                if (AskIt(filtered))
+                {
+                    if (AskWithWarningsForRequiredBundles(filtered))
+                    {
+                        DoIt(filtered.Keys);
+                    }
+                }
             }
         }
 
@@ -201,25 +207,43 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
             return args;
         }
 
-        private static void AskIt(IEnumerable<Bundle> bundles)
+        private static bool AskIt(IDictionary<Bundle, string> bundles)
         {
-            var displayNames = string.Join("\n", bundles.Select(bundle => $"  {bundle.DisplayName}"));
+            var displayNames = string.Join("\n", bundles.Select(bundle => $"  {bundle.Key.DisplayName}"));
             Console.Write(string.Format(LocalizableStrings.ConfirmationPromptOutputFormat, displayNames));
 
             var response = Console.ReadLine().Trim().ToUpper();
 
-            if (response.Equals("Y"))
+            if (response.Equals("Y") || response.Equals("N"))
             {
-                DoIt(bundles);
-            }
-            else if (response.Equals("N"))
-            {
-                return;
+                return response.Equals("Y");
             }
             else
             {
                 throw new ConfirmationPromptInvalidException();
             }
+        }
+
+        private static bool AskWithWarningsForRequiredBundles(IDictionary<Bundle, string> bundles) 
+        {
+            var requiredBundles = bundles.Where(b => !b.Value.Equals(string.Empty));
+            foreach (var pair in requiredBundles)
+            {
+                Console.ForegroundColor = ConsoleColor.Red;
+                Console.Write(string.Format(LocalizableStrings.RequiredBundleConfirmationPromptOutputFormat, pair.Key.DisplayName, pair.Value));
+                Console.ResetColor();
+                var response = Console.ReadLine().Trim().ToUpper();
+                if (response.Equals("N"))
+                {
+                    return false ;
+                }
+                else if (!response.Equals("Y"))
+                {
+                    throw new ConfirmationPromptInvalidException();
+                }
+            }
+
+            return true;
         }
     }
 }
--- a/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
+++ b/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
@@ -4,9 +4,6 @@ using Microsoft.DotNet.Tools.Uninstall.Shared.Configs;
 using Microsoft.DotNet.Tools.Uninstall.Shared.Exceptions;
 using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
 using Microsoft.DotNet.Tools.Uninstall.Shared.Utils;
-using Microsoft.DotNet.Tools.Uninstall.Windows;
-using System.Reflection;
-using Microsoft.DotNet.Tools.Uninstall.MacOs;
 using System.Linq;
 using System.Diagnostics;
 using Microsoft.DotNet.Tools.Uninstall.Shared.Configs.Verbosity;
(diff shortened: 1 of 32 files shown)


wli3:
Give me sometime to review this, quite big one

sfoslund:
@wli3 No problem, I will be pushing changes later anyway (nothing big, just rewording help language and such)

wli3:
"whatif" is the same as "dry-run" right? The current help is like the following, duplicated help is confusing. Maybe the description of "whatif" should just be "alias of dry-run".

```
dotnet-core-uninstall:
  Remove specified .NET Core SDKs or Runtimes. This tool can only uninstall items that were installed using .NET Core SDK or Runtime installers. Read the documentation of this tool at https://aka.ms/dotnet-core-uninstall.

Usage:
  dotnet-core-uninstall [options] [command]

Options:
  --version    Display version information

Commands:
  list                 List .NET Core SDKs or Runtimes that can be removed with this tool.
  dry-run <VERSION>    Display .NET Core SDKs and Runtimes that will be removed.
  whatif <VERSION>     Display .NET Core SDKs and Runtimes that will be removed.
  remove <VERSION>     Remove the specified .NET Core SDKs or Runtimes.
```

src/dotnet-core-uninstall/Shared/Configs/CommandLineConfigs.cs
```diff
@@ -171,74 +167,57 @@ internal static class CommandLineConfigs
...
+        static CommandLineConfigs() 
         {
-            UninstallRootCommand.AddArgument(new Argument<IEnumerable<string>>
-            {
-                Name = LocalizableStrings.UninstallNoOptionArgumentName,
-                Description = LocalizableStrings.UninstallNoOptionArgumentDescription
-            });
-
             UninstallRootCommand.AddCommand(ListCommand);
+            UninstallRootCommand.AddCommand(DryRunCommand);
```
wli3: I find system.commandline has build-in alias for command. 

Related to the comment below for string vs Option. We do need to use this field later. So it does not need to be static then, make it readonly and initialize in constructor. 

```c#
// put in constructor
            this.DryRunCommand = new Command(
               DryRunCommandName,
               LocalizableStrings.DryRunCommandDescription);

            DryRunCommand.AddAlias("whatif");

// put in constructor end
            UninstallRootCommand.AddCommand(DryRunCommand);
```

src/dotnet-core-uninstall/Shared/Configs/OptionFilterers.cs
```diff
@@ -1,44 +1,43 @@
...
-using System.CommandLine;
 using Microsoft.DotNet.Tools.Uninstall.Shared.Filterers;
 
 namespace Microsoft.DotNet.Tools.Uninstall.Shared.Configs
 {
     internal static class OptionFilterers
     {
-        public static readonly IDictionary<Option, Filterer> OptionFiltererDictionary
-            = new Dictionary<Option, Filterer>
+        public static readonly IDictionary<string, Filterer> OptionFiltererDictionary
```
wli3: Do we have to use string here? Option is a stronger type and can have more type safety (and also better "documentation" considering it is clear only a certain type can put in the dictionary instead of any string)

test/dotnet-core-uninstall.Tests/Shared/Configs/CommandLineConfigsTests.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace Microsoft.DotNet.Tools.Uninstall.Tests.Shared.Configs
 {
-    public class CommandLineConfigsTests
+    public class CommandLineConfigsTests
```
wli3: nit: white spaces. Usually run a git diff (extra white space will be red) before git add --all can help prevent this problem

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -14,9 +16,12 @@ internal static class VisualStudioSafeVersionsExtractor
         // Divisions within version bands. Not inclusive, groupings: [2.1.000, 2.1.600) [2.1.600, 2.2.00) [2.0.00, 2.2.200) [2.2.200, 2.3.00)
         private static readonly SemanticVersion[] SpecialCaseDivisions = { new SemanticVersion(2, 1, 600), new SemanticVersion(2, 2, 200) };
 
+        // The tool should not be used to uninstall any more recent versions of the sdk
+        private static readonly SemanticVersion UpperLimit = new SemanticVersion(3, 0, 0);
```
wli3: @KathleenDollard should we also put this somewhere in help to also let the user know why this tool cannot remove 3.0 (post 3.0 it is handled properly by VS)

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -48,5 +53,28 @@ private static IEnumerable<IEnumerable<Bundle>> DivideSpecialCases(IEnumerable<B
...
+        public static Dictionary<Bundle, string> GetListCommandUninstallableStrings(IEnumerable<Bundle> allBundles)
+        {
+            var uninstallable = GetUninstallableBundles(allBundles);
+            var required = allBundles.Where(b => !uninstallable.Contains(b));
+
+            var ListCommandStringResults = uninstallable.Select(b => new KeyValuePair<Bundle, string>(b, string.Empty))
+                        .ToDictionary(i => i.Key, i => i.Value);
+            if (required.Where(b => b.Version.SemVer < SpecialCaseDivisions[0]).Count() > 0) 
+            {
+                ListCommandStringResults.Add(required.Where(b => b.Version.SemVer < SpecialCaseDivisions[0]).Max(), "[Required by Visual Studio 2017]");
```
wli3: [Required by Visual Studio 2017] part need loc as well

src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
```diff
@@ -201,25 +207,43 @@ private static IEnumerable<string> ParseCommandToArgs(string command)
...
+        private static bool AskIt(IDictionary<Bundle, string> bundles)
         {
-            var displayNames = string.Join("\n", bundles.Select(bundle => $"  {bundle.DisplayName}"));
+            var displayNames = string.Join("\n", bundles.Select(bundle => $"  {bundle.Key.DisplayName}"));
             Console.Write(string.Format(LocalizableStrings.ConfirmationPromptOutputFormat, displayNames));
 
             var response = Console.ReadLine().Trim().ToUpper();
 
-            if (response.Equals("Y"))
+            if (response.Equals("Y") || response.Equals("N"))
```
wli3: nit: Even though this change is shorter, but not as easier to read as before

src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
```diff
@@ -201,25 +207,43 @@ private static IEnumerable<string> ParseCommandToArgs(string command)
             return args;
         }
 
-        private static void AskIt(IEnumerable<Bundle> bundles)
+        private static bool AskIt(IDictionary<Bundle, string> bundles)
```
wli3: nit, good to improve but not blocking: the return value of AskIt is not obvious, it is the user's answer? or it is if this operation succeed or not.

A quick fix is to Rename it "AskItAndReturnUserAnswer" or simply add documentation.
A bigger design style is to make side effects(print to screen) API be void, and compute (query) only API to be return only and side effects free. So let it return a value might not be ideal
https://blog.ploeh.dk/2015/10/08/command-query-separation-when-queries-should-have-side-effects/

wli3:
Awesome work Sara!

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -48,5 +53,28 @@ private static IEnumerable<IEnumerable<Bundle>> DivideSpecialCases(IEnumerable<B
...
+            }
+            foreach (var recentSdk in required.Where(b => b.Version.SemVer >= UpperLimit))
+            {
+                ListCommandStringResults.Add(recentSdk, $"[Cannot uninstall version {UpperLimit} and above]");
```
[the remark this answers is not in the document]
sfoslund: I'm going to talk more to Kathleen about this when she is back tomorrow, since I agree that it would be helpful to have more information about why SDKs aren't allowed to be uninstalled but we don't want these strings to create clutter in the list command output. For now I've added more information when verbosity is set to anything above normal.

sfoslund:
@wli3 I've finished making additional changes to this PR, thanks for your patience!

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -8,76 +7,88 @@
...
-    //                        16.0 requires anything less than 3.0.1xx
-    // Must keep one of each Major.Minor band to ensure runtime works
     internal static class VisualStudioSafeVersionsExtractor
     {
-        // Divisions within version bands. Not inclusive, groupings: [2.1.000, 2.1.600) [2.1.600, 2.2.00) [2.0.00, 2.2.200) [2.2.200, 2.3.00)
-        private static readonly SemanticVersion[] SpecialCaseDivisions = { new SemanticVersion(2, 1, 600), new SemanticVersion(2, 2, 200) };
-
         // The tool should not be used to uninstall any more recent versions of the sdk
-        public static readonly SemanticVersion UpperLimit = RuntimeInfo.RunningOnWindows ? new SemanticVersion(3, 0, 0) : new SemanticVersion(5, 0, 0);
+        public static readonly SemanticVersion UpperLimit = new SemanticVersion(5, 0, 0);
```
wli3: we change it from 3.0 to 5.0 why?will it try to uninstall the dummy entry Visual Studio created?

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -2,51 +2,93 @@
...
+                    dividedBundles.Add(bundlesInRange, explaination);
+                }
+            }
+
+            return (dividedBundles, bundleList);
+        }
 
         public static IEnumerable<Bundle> GetUninstallableBundles(IEnumerable<Bundle> bundles)
         {
+            if (!RuntimeInfo.RunningOnWindows) // TODO this will be changed when mac protection is added
```
wli3: this is enough for mac protection right? can we remove it now?

in general no TODO in code, you can file an issue and link the code there

wli3: also, if you feel you really need to add a note to tell where to add the next feature, maybe it is an indication to refactor the code to make it more modular so the place to change is obvious.

wli3:
I am worried about uninstalling the dummy entry created by Visual Studio. We dodged this problem by cap to 3.0 (only 3.0 starts to have this dummy entry), although i don't know if we have other code to avoid it checked in already.

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
sfoslund: The upper limit was mostly meant to cap the use of the tool so people don't continue to use it to uninstall future versions. There's still logic to protect the highest version below that upper limit (usually 3.1.xxx or 3.0.xxx) [here](https://github.com/sfoslund/cli-lab/blob/RestructureCommands/src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs#L25). I'm not sure what the Visual Studio dummy entry is, can you give me some more information?

wli3:
Just double check with John on the dummy entry issue. And this is good to merge

PR #97, opened 2019-10-22
Visual Studio for Mac version protection

sfoslund:
Adding protection to ensure the tool does not uninstall bundles required by Visual Studio for Mac. Prevents (without the --force option) the user from uninstalling the highest installed major.minor runtime versions and the highest installed sdk version.

### DIFF
--- a/src/dotnet-core-uninstall/Shared/Configs/CommandLineConfigs.cs
+++ b/src/dotnet-core-uninstall/Shared/Configs/CommandLineConfigs.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.CommandLine;
+using System.CommandLine.Builder;
 using System.CommandLine.Invocation;
 using System.Linq;
 using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
@@ -227,7 +228,9 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Configs
             DryRunCommand.Handler = CommandHandler.Create(ExceptionHandler.HandleException(() => DryRunCommandExec.Execute()));
             RemoveCommand.Handler = CommandHandler.Create(ExceptionHandler.HandleException(() => UninstallCommandExec.Execute()));
 
-            CommandLineParseResult = UninstallRootCommand.Parse(Environment.GetCommandLineArgs());
+            var parser = new CommandLineBuilder(UninstallRootCommand)
+                         .Build();
+            CommandLineParseResult = parser.Parse(Environment.GetCommandLineArgs());
         }
 
         public static Option GetUninstallMainOption(this CommandResult commandResult)
--- a/src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
+++ b/src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
@@ -49,13 +49,13 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.VSVersioning
             var dividedBundles = bundleList
                 .Where(bundle => bundle.Version is RuntimeVersion)
                 .GroupBy(bundle => bundle.Version.MajorMinor)
-                .Select(pair => (pair as IEnumerable<Bundle>, string.Format(LocalizableStrings.RequirementExplainationString, " or SDKs")))
+                .Select(pair => (pair as IEnumerable<Bundle>, string.Format(LocalizableStrings.RequirementExplainationString, " for Mac or SDKs")))
                 .ToDictionary(key => key.Item1, value => value.Item2); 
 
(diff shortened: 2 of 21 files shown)


src/dotnet-core-uninstall/Shared/Configs/CommandLineConfigs.cs
```diff
@@ -227,7 +228,9 @@ static CommandLineConfigs()
             DryRunCommand.Handler = CommandHandler.Create(ExceptionHandler.HandleException(() => DryRunCommandExec.Execute()));
             RemoveCommand.Handler = CommandHandler.Create(ExceptionHandler.HandleException(() => UninstallCommandExec.Execute()));
 
-            CommandLineParseResult = UninstallRootCommand.Parse(Environment.GetCommandLineArgs());
+            var parser = new CommandLineBuilder(UninstallRootCommand)
```
wli3: Why this change? `CommandLineBuilder(UninstallRootCommand)` should be the default.

test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
```diff
@@ -64,97 +77,167 @@ internal void TestGetUninstallable(string[] versions, bool[] allowed)
...
-        [InlineData(new string[] { "5.0.0", "5.0.1", "10.100.100" }, new bool[] { true, true, true })]
-        internal void TestGetUninstallableNonSdkVersionsMac(string[] versions, bool[] allowed)
-        {
-            TestGetUninstallableNonSdkVersions(versions, allowed);
-        }
-
-        internal void TestGetUninstallableNonSdkVersions(string[] versions, bool[] allowed)
+        [InlineData(new string[] { "1.0.0" }, new bool[] { false }, new string[] { "1.0.0" }, new bool[] { false })]
+        [InlineData(new string[] { "1.0.0", "1.0.1" }, new bool[] { true, false }, new string[] { "1.0.0", "1.0.1" }, new bool[] { true, false })]
+        [InlineData(new string[] { "2.1.0", "1.0.1" }, new bool[] { false, true }, new string[] { "2.0.0", "1.1.0" }, new bool[] { false, false })]
```
wli3: new string[] { "2.1.0", "1.0.1" }, new bool[] { false, true }

why 2.1.0 is allowed but 1.01 is not? Did we only want 3.0.0 for VS for mac?

src/dotnet-core-uninstall/Shared/Configs/CommandLineConfigs.cs
sfoslund: I'm working on customizing the help builder right now which will require a parser build from the command line builder. I didn't mean to include this change in this commit, but it will be part of the next PR.

test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
sfoslund: VS for mac only needs the highest SDK, so we only ever have to keep one. That test is saying that we should be able to uninstall 1.0.1 (it is allowed to be uninstalled) but 2.1.0 is required by VS for mac since it's the highest SDK "installed"

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -41,13 +41,43 @@ private static (IDictionary<IEnumerable<Bundle>, string>, IEnumerable<Bundle>) A
...
+        private static (IDictionary<IEnumerable<Bundle>, string>, IEnumerable<Bundle>) ApplyMacVersionDivisions(IEnumerable<Bundle> bundleList)
         {
-            if (!RuntimeInfo.RunningOnWindows)
+            var bundlesAboveLimit = bundleList.Where(bundle => bundle.Version.SemVer >= UpperLimit);
+            bundleList = bundleList.Except(bundlesAboveLimit);
+
+            var dividedBundles = bundleList
+                .Where(bundle => bundle.Version is RuntimeVersion)
+                .GroupBy(bundle => bundle.Version.MajorMinor)
+                .Select(pair => (pair as IEnumerable<Bundle>, string.Format(LocalizableStrings.RequirementExplainationString, " for Mac or SDKs")))
```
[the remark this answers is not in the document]
sfoslund: Not totally sure what you mean... Are you suggesting using a Mac specific version of the RequirementExplainationString? That seems like a good idea.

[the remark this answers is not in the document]
wli3: "Maybe needed for Visual Studio" + " for Mac or SDKs" when translated separately to different languages and then join together probably won't all make sense.

Might just have "Maybe needed for Visual Studio for Mac or SDKs" as one string

Issue #109, opened 2020-01-06
`--all-previews` does not uninstall all preview SDKs


sfoslund:
From manual testers: On mac

Setup:
Stable SDK + Multi Preview SDK 

--all-previews will not remove all preview SDKs, the latest Preview SDK need to use --force to be removed. 

![GetImage (6)](https://user-images.githubusercontent.com/16010855/71840583-707f0900-3072-11ea-8fb5-bf1d3ddf451b.png)

sfoslund:
Fixed by #111

Issue #108, opened 2020-01-06
Lower patch cannot be uninstalled without `--force`


sfoslund:
From manual testers: On Mac

Setup:
- Install VSFM8.3 on stable channel ( SDK 3.0.101 + Runtime 3.0.1 & 2.1.14 installed), Then update VSFM to 8.4 Preview(SDK 3.1.100 + Runtime 3.1.0 installed) 
- Manually install higher patch version SDK 3.1.101(Runtime 3.1.1) 

The lower patch 3.1.100 could only be removed using --force

![GetImage (5)](https://user-images.githubusercontent.com/16010855/71840395-1da55180-3072-11ea-94a5-f842ecf08cd2.png)

sfoslund:
Fixed by #111

Issue #106, opened 2020-01-06
ASP.NET Core Runtimes could not be displayed


sfoslund:
From manual testers: On Windows

The ASP.NET Core Runtimes info could not be displayed 
- If Installed ASP.NET Core Runtime 3.0.0-preview7/8/9/rc1 
- If Installed ASP.NET Core Runtime 3.1.0-preview1/2/3 
Note: This issue is not repro with ASP.Net Core Runtime 3.0.0-preview1/2/3/4/5/6 

![GetImage (3)](https://user-images.githubusercontent.com/16010855/71840054-6dcfe400-3071-11ea-8de7-058832ec6720.png)

yuki233:
The ASP.Net Core Runtime installers could be download using below URL
https://dotnet.microsoft.com/download/dotnet-core/3.0
https://dotnet.microsoft.com/download/dotnet-core/3.1

sfoslund:
Fixed by #111

PR #111, opened 2020-01-13
Fixing Issues from manual testing

### LINKED ISSUE #103: List: specified version not valid error
From manual testers: On Windows

List will return error -The specified version is not valid: ""  
- If installed 3.0.0 or 3.1.0 ASP.Net Core Runtime separately. 
- If installed 3.0.0-rc1 or 3.1.0-preview x .Net Core Runtime separately 

Notes:
- There is no ASP.Net Core Runtime installer for Mac, so this issue is not occurred on Mac. 
- If installed 3.0.0-rc1 or 3.1.0-preview x .Net Core Runtime separately for Mac, it is not repro on Mac. 

GetImage

### LINKED ISSUE #105: `--version` option fails on `dotnet-core-uninstall`
From manual testers: On Windows

Run "dotnet-core-uninstall",  the option " --version" doesn't work. 

GetImage (2)

### LINKED ISSUE #107: `--x86` option does not display all x86 SDKs
From manual testers: On Windows

Run "dotnet-core-uninstall dry-run --all --sdk", the option " --x86" doesn't work. Only shows one x86 SDK. 

GetImage (4)

sfoslund:
- Restructuring to more oo style for ease of testing
- Adding integration tests

Issues addressed:
- Fixes #103, #106: Fix ASP.NET core runtime regex parsing
- Fixes #107, #108, #109: Fix input of VS version protection calls to ensure consistency
- Fixes #105: Fix command line structure to support `--version` option


src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
```diff
@@ -215,7 +216,7 @@ private static bool AskItAndReturnUserAnswer(IDictionary<Bundle, string> bundles
 
             var response = Console.ReadLine().Trim().ToUpper();
 
-            if (response.Equals("Y"))
+            if (response.Equals("Y") || response.Equals("YES"))
```
wli3: Suggestion: We could dependency inject `Console.ReadLine().Trim().ToUpper();`, or extract method to make the `response` a pass-in parameter. So we can also test Y and Yes are acceptable answer.

src/dotnet-core-uninstall/Shared/Commands/ListCommandExec.cs
```diff
@@ -8,33 +8,25 @@
...
-            {
-                Execute(
-                    FileSystemExplorer.GetAllInstalledBundles(),
-                    MacOs.SupportedBundleTypeConfigs.SupportedBundleTypes);
-            }
-            else
-            {
-                throw new OperatingSystemNotSupportedException();
-            }
+            SysConsole = console == null? new SystemConsole() : console;
```
wli3: suggestion: Looks like there are multiple bugs. It is better to have separate commits with tests per bug, so the viewer (and reader after it is checked in) know which bug corresponding to which test.

If it is hard to separate. You can just add some pull request comments to point them out. Like "this line is for bug A" "this line is the test for bug A"

wli3: And here i really don't know if it is a refactoring about SysConsole init or it is a bug

wli3: Same for refactoring and feature mix. Ideally refactoring commit should be separated with feature/bug fix. Unless it is obvious to separate them out like they are in different files.

If it hard to separate, it's better to add some comment in PR.

sfoslund:
@wli3 I'll add some clarification on which tests are for which bugs, my reasoning behind some of the refactoring, etc tomorrow morning if you want to hold off on reviewing until then.

test/dotnet-core-uninstall.Tests/Shared/IntegrationTests/IntegrationTests.cs
```diff
@@ -0,0 +1,128 @@
...
+using FluentAssertions;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Commands;
+using Microsoft.DotNet.Tools.Uninstall.Shared.Configs;
+using Microsoft.DotNet.Tools.Uninstall.Tests.Attributes;
+using Microsoft.DotNet.Tools.Uninstall.Windows;
+using Xunit;
+
+namespace Microsoft.DotNet.Tools.Uninstall.Tests.Shared.IntegrationTests
+{
+    public class IntegrationTests
```
wli3: Suggestion: The "abstraction level" of this integration test is not good. And in general you should reduce integration test coverage and replace with unit test coverage. The downside of integration tests are (I can think of so far)

1. slow
2. flaky
3. test itself is complicated
4. hard to understand why it breaks
5. does not encourage better abstraction
6. combinations of features require exponential integration tests to cover

to be fair, the upside is, 

1. if the abstraction layer is right, it is easy for the reader not familiar with the code to understand the test. Like when you can test CLI by passing command line argument and assert the screen output.
2. You could have less tests to cover a whole stack, and if you do refactoring, you don't need to edit test. (at the same time, if your tests' level of abstraction is wrong, a small change could cost you to edit all tests, back to the example above, if you changed the output format, out assertion need to be changed)
3. Cover integration between "units" easily. (Although it is still possible to test things like dependency injection in unit test)

So a large number of people prefer integration tests. And in reality, a lot of cases are just too hard to put under unit test.

For this specific case, I think the problem is 3, 4. It is complicated in method `ListCommandExecutionIsCorrect` which means the test itself could be wrong. And for 4, I really cannot tell which part you want to test and if you reproduced the bug in the tests.

I won't block the PR, but you should keep the trade off in mind.

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
```diff
@@ -40,65 +42,81 @@ public static IEnumerable<Bundle> GetAllInstalledBundles()
...
 
-        private static void ParseVersionAndArch(string displayName, string bundleCachePath, out BundleVersion version, out BundleArch arch)
+        private void ParseVersionAndArch(RegistryKey registryKey, string displayName, string bundleCachePath, out BundleVersion version, out BundleArch arch)
         {
             var match = Regexes.BundleDisplayNameRegex.Match(displayName);
             var cachePathMatch = Regexes.BundleCachePathRegex.Match(bundleCachePath);
             var archString = cachePathMatch.Groups[Regexes.ArchGroupName].Value ?? string.Empty;
-            var versionString = cachePathMatch.Groups[Regexes.VersionGroupName].Value;
+            var versionFromCachePath = cachePathMatch.Groups[Regexes.VersionGroupName].Value;
+            var versionFromRegistry = string.Join('.', (registryKey.GetValue("DisplayVersion") as string).Split('.').Take(3));
```
wli3: could you help me understand this logic? Maybe add some comment, why we can also get the result from CachePath

sfoslund: Sure, this is essentially the underlying problem with #103 and #106. The cache path for asp.net runtimes doesn't have the version in it so the version can't be parsed out in the same way that it is for the other bundle types. This was causing an error when you tried to `list` on a machine with these runtimes, since it couldn't be parsed in the expected way. 
I'm fixing this here by getting the version from the registry, since this is a more reliable way to get the version for asp.net runtimes.
Are you suggesting adding a comment in the code about this?

wli3: Yes, it is a good place to add comment. It is an external system's behavior and it is not well known at all

wli3: btw there is a way to "document" by test, but it is over kill in this case

Blog post from a friend of mine
https://docs.microsoft.com/en-us/archive/blogs/ericgu/portadaptersimulator-read-only-and-write-only-dependencies

test/dotnet-core-uninstall.Tests/Shared/IntegrationTests/IntegrationTests.cs
sfoslund: Thanks for the explanation, that makes sense. The main 2 bugs I found were issues with how methods were called in high level functions (I believe execute functions for some of the commands) so I wasn't sure how to test them, especially with the functional style the tool is written in.
I've refactored a bit in the last update so I can cover the bugs without the confusing integration tests- let me know what you think.

sfoslund:
@wli3 I made some nontrivial changes based on your feedback about the integration tests- am I still okay to merge this or do you have any other feedback?

wli3:
Reviewed again, still good to go

PR #170, opened 2021-01-05
Add ability to uninstall 5.0

sfoslund:
Tested manually on windows and mac

### DIFF
--- a/src/dotnet-core-uninstall/Shared/Utils/Regexes.cs
+++ b/src/dotnet-core-uninstall/Shared/Utils/Regexes.cs
@@ -19,25 +19,17 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Utils
             $@"(?<{MajorGroupName}>\d+)\.(?<{MinorGroupName}>\d+)");
         private static readonly Regex _previewVersionNumberRegex = new Regex(
             $@"(\d+(\.\d+)*)?");
-        private static readonly Regex _previewVersionNumberRuntimeDisplayNameRegex = new Regex(
-            $@"(\s\d+(\.\d+)?)?");
         private static readonly Regex _rcVersionNumberRegex = new Regex(
             $@"\d+");
         private static readonly Regex _buildNumberRegex = new Regex(
             $@"(?<{BuildGroupName}>\d+)");
         private static readonly Regex _archRegex = new Regex(
-            $@"(?<{ArchGroupName}>x64|x86)");
+            $@"(?<{ArchGroupName}>\-?x64|x86)");
 
         private static readonly Regex _previewVersionSdkDisplayNameRegex = new Regex(
-            $@"(?<{PreviewGroupName}>\s\-\s(preview{_previewVersionNumberRegex.ToString()}|rc{_rcVersionNumberRegex.ToString()}))");
-        private static readonly Regex _previewVersionRuntimeDisplayNameRegex = new Regex(
-            $@"(?<{PreviewGroupName}>\s(Preview{_previewVersionNumberRuntimeDisplayNameRegex.ToString()}|Release\sCandidate\s{_rcVersionNumberRegex.ToString()}))");
-        private static readonly Regex _previewVersionAspNetRuntimeDisplayNameRegex = new Regex(
-            $@"(?<{PreviewGroupName}>\s(Preview{_previewVersionNumberRuntimeDisplayNameRegex.ToString()}(\sBuild\s{_buildNumberRegex.ToString()}(\.\d+|\-\d+)?)?|Release\sCandidate\s{_rcVersionNumberRegex.ToString()}))");
-        private static readonly Regex _previewVersionHostingBundleDisplayNameRegex = new Regex(
-            $@"(?<{PreviewGroupName}>\s(Preview{_previewVersionNumberRuntimeDisplayNameRegex.ToString()}(\s(Build\s{_buildNumberRegex.ToString()}(\.\d+|\-\d+)?|Build\spreview{_previewVersionNumberRegex.ToString()}|{_buildNumberRegex.ToString()}(\s{_buildNumberRegex.ToString()})?(\spreview{_previewVersionNumberRegex.ToString()})?))?|Release\sCandidate\s{_rcVersionNumberRegex.ToString()}))");
+            $@"(?<{PreviewGroupName}>\s?\-\s?((preview|alpha)\.?{_previewVersionNumberRegex.ToString()}|rc{_rcVersionNumberRegex.ToString()}))");
         private static readonly Regex _previewVersionSdkCachePathRegex = new Regex(
-            $@"(?<{PreviewGroupName}>\-(preview{_previewVersionNumberRegex.ToString()}|rc{_rcVersionNumberRegex.ToString()}(\.\d+)?)\-(?<{BuildGroupName}>\d+))");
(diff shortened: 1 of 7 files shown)


joeloff:
Looks good. I noticed there's some code that examines the DisplayVersion properties from the registry. Because of how the actual versions are calculated, the version in the DisplayName and the DisplayVersion will be different for 5.x installs. For example, for 5.0.100-preview.2-20176.6, the actual file version (which would also be used for the DisplayVersion) will be 5.1.20.17606.

Not sure if that impacts what the tool expects to receive as input, but just wanted to call it out.

joeloff (approved):

sfoslund:
> the DisplayVersion will be different for 5.x installs

Yeah, I noticed that so I removed some code that was using DisplayVersion before. I think the only uses now are taken directly from CoreScan and are just checking that it's not empty- we can always assume it's non-empty, right?

joeloff:
> Yeah, I noticed that so I removed some code that was using DisplayVersion before. I think the only uses now are taken directly from CoreScan and are just checking that it's not empty- we can always assume it's non-empty, right?

Correct. The DisplayVersion should always be set.

src/dotnet-core-uninstall/Shared/Utils/Regexes.cs
```diff
@@ -53,29 +45,6 @@ internal static class Regexes
...
-        private static readonly Regex _runtimeVersionDisplayNameRegex = new Regex(string.Format(
-            _runtimeVersionBasicRegexFormat,
-            _previewVersionRuntimeDisplayNameRegex.ToString()));
-        private static readonly Regex _aspNetRuntimeVersionDisplayNameRegex = new Regex(string.Format(
-            _runtimeVersionBasicRegexFormat,
-            _previewVersionAspNetRuntimeDisplayNameRegex.ToString()));
-        private static readonly Regex _hostingBundleVersionDisplayNameRegex = new Regex(string.Format(
-            _runtimeVersionBasicRegexFormat,
-            _previewVersionHostingBundleDisplayNameRegex.ToString()));
-        private static readonly Regex _hostingBundleAuxVersionDisplayNameRegex = new Regex(string.Format(
```
wli3, line 68: these are removed due to applying corescan logic?

test/dotnet-core-uninstall.Tests/Shared/Utils/RegexesTests.cs
```diff
@@ -68,103 +47,6 @@ internal void TestBundleMajorMinorRegexReject(string input)
...
-        [InlineData(".NET Core SDK 1.1.10 (x64)")]
-        [InlineData("Microsoft .NET Core SDK - 2.1.4 (x86)")]
-        [InlineData("Microsoft .NET Core Runtime - 2.2.4 (x64)")]
-        [InlineData("Microsoft .NET Core Runtime - 2.1.0 Release Candidate 1 (x86)")]
-        [InlineData("Microsoft .NET Core Runtime - 3.0.0 Preview 5 (x64)")]
-        [InlineData("Microsoft .NET Core 1.1.13 - Runtime (x64)")]
-        [InlineData("Microsoft ASP.NET Core 3.0.0 Preview 6 Build 19307.2 - Shared Framework")]
-        [InlineData("Microsoft ASP.NET Core 2.2.6 - Shared Framework")]
-        [InlineData("Microsoft ASP.NET Core 2.2.0 Preview 3 Build 35497 - Shared Framework")]
-        [InlineData("Microsoft ASP.NET Core 2.1.0 Release Candidate 1 - Shared Framework")]
```
wli3, line 84: why these tests are removed? due to corescan logic does not support it?

wli3 (commented):
could you point out which part is port from corescan? So I don't need to review that part

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
```diff
@@ -49,36 +50,34 @@ public virtual IEnumerable<Bundle> GetAllInstalledBundles()
...
-                false :
-                Regexes.BundlePublisherRegex.IsMatch(publisher);
+            return IsNetCoreBundle(uninstallKey.GetValue("DisplayName") as string,
+                uninstallKey.GetValue("DisplayVersion") as string,
+                uninstallKey.GetValue("UninstallString") as string,
+                uninstallKey.GetValue("BundleVersion") as string);
         }
 
-        private static bool IsDotNetCoreBundleUninstaller(int? windowsInstaller)
+        internal static bool IsNetCoreBundle(string displayName, string displayVersion, string uninstallString, string bundleVersion)
```
joeloff, line 66: @wli3 Most of this method comes from CS

src/dotnet-core-uninstall/Shared/Utils/Regexes.cs
sfoslund: Yes, they're no longer needed once we convert to corescan

test/dotnet-core-uninstall.Tests/Shared/Utils/RegexesTests.cs
sfoslund: Half of the deleted tests are moved [here](https://github.com/dotnet/cli-lab/pull/170/files#diff-b36d4685ade1f3d7383ef3ff9de9d9233eb115cec27950ba348518fc290c4ed5R13) since they are no longer directly testing the regex. The other half are removed because corescan's logic is less restrictive about which bundles are considered .NET bundles. With the regex's we were looking for very specific display name strings and with the corescan logic we're less exact.

For example, out original logic might not have marked `Microsoft ASP.NET Core 2.1.0 Release Candidate 1 - Shared Framework` as a .NET bundle because there are extra spaced around the `-`. The corescan logic doesn't care about that kind of thing, which is good because it will be more resillient to small format changes.

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
```diff
@@ -97,41 +97,59 @@ private static Bundle WrapRegistryKey(RegistryKey registryKey)
...
-            // Note: ASP.NET Core runtimes do not include version in the cache path, need to get version from registry:
-            var versionFromRegistry = string.Join('.', (registryKey.GetValue("DisplayVersion") as string).Split('.').Take(3));
-            var versionString = string.IsNullOrEmpty(versionFromCachePath) ? versionFromRegistry : versionFromCachePath;
             var hasAuxVersion = cachePathMatch.Groups[Regexes.AuxVersionGroupName].Success;
             var footnote = hasAuxVersion ?
                 string.Format(LocalizableStrings.HostingBundleFootnoteFormat, displayName, versionString) :
                 null;
 
-            if (string.IsNullOrEmpty(displayName) || string.IsNullOrEmpty(versionString))
+            // Classify the bundle type
```
sfoslund, line 109: This section is also taken from corescan

sfoslund:
@wli3 the 2 main sections that were taken from corescan were marked above: 
https://github.com/dotnet/cli-lab/pull/170/files#diff-bad95d57529fabfa7c211b95dca8e77cf94d58267bf9ddde1746e4130e4009dfR109 and https://github.com/dotnet/cli-lab/pull/170/files#diff-bad95d57529fabfa7c211b95dca8e77cf94d58267bf9ddde1746e4130e4009dfR53

wli3 (approved):

Issue #160, opened 2020-09-13
How to uninstall the preview version SDKs ？


cnbohu:
1.0.0-preview1-002702 [C:\Program Files\dotnet\sdk]
1.0.0-preview2-003131 [C:\Program Files\dotnet\sdk]
1.0.0-preview2-003156 [C:\Program Files\dotnet\sdk]

sfoslund:
Are these previews not coming up when you run `dotnet-core-uninstall list`?

DanielMeixner:
I have a similar issue, in my case with 5.0-preview. 


**dotnet --list-sdks**
__2.1.202 [C:\Program Files\dotnet\sdk]
2.1.508 [C:\Program Files\dotnet\sdk]
2.1.509 [C:\Program Files\dotnet\sdk]
2.1.511 [C:\Program Files\dotnet\sdk]
2.1.512 [C:\Program Files\dotnet\sdk]
2.1.513 [C:\Program Files\dotnet\sdk]
2.1.515 [C:\Program Files\dotnet\sdk]
2.1.517 [C:\Program Files\dotnet\sdk]
2.1.518 [C:\Program Files\dotnet\sdk]
2.1.801 [C:\Program Files\dotnet\sdk]
3.1.402 [C:\Program Files\dotnet\sdk]
3.1.403 [C:\Program Files\dotnet\sdk]_
5.0.100-preview.8.20417.9 [C:\Program Files\dotnet\sdk]_

**dotnet-core-uninstall list**

__This tool can not uninstall versions of the runtime or SDK that are 
    - SDKs installed using Visual Studio 2019 Update 3 or later.
    - SDKs and runtimes installed via zip/scripts.
    - Runtimes installed with SDKs (these should be removed by removing that SDK).
The versions that can be uninstalled with this tool are:_

_.NET Core SDKs:
  3.1.402  x64    [Used by Visual Studio. Specify individually or use --force to remove]
  2.1.801  x64    [Used by Visual Studio 2019. Specify individually or use --force to remove]
  2.1.518  x64    [Used by Visual Studio 2017. Specify individually or use --force to remove]
  2.1.517  x64
  2.1.515  x64
  2.1.513  x64
  2.1.512  x64
  2.1.511  x64
  2.1.509  x64
  2.1.508  x64
  2.1.202  x64    [Used by Visual Studio. Specify individually or use --force to remove]__

[...]

PR #195, opened 2021-09-02
Add support for .NET 6.0

sfoslund:
Fixes https://github.com/dotnet/cli-lab/issues/160


joeloff (approved):

wli3:
Ideally we should also run uninstall for all workloads first so nothing left on the disk

joeloff:
If you uninstall the SDK it will clean up workloads (the SDK bundle).

Issue #186, opened 2021-06-28
dotnet-core-uninstall unable to list `Hosting Bundle` 5.0.x


yyjdelete:
1. After install ASP.NET Core Runtime 5.0.x(Hosting Bundle for windows), it's not listed in `.NET Core Runtime & Hosting Bundles:` sector. The same thing works well for 3.1.x;

sfoslund:
@yyjdelete are you still seeing this issue? If so, can you provide the following information:

1. Your OS
2. The version of the uninstall tool you're using
3. The full version of the hosting bundle that's failing to show up

Thanks!

yyjdelete:
Yes, it still not work for latest version

Your OS: win10-x64(21H1, 19043.1110)
The version of the uninstall tool you're using: 1.2.206301
The full version of the hosting bundle that's failing to show up: 5.0.9
https://download.visualstudio.microsoft.com/download/pr/a0f49856-eec9-4962-8d81-b09af6be9435/1d5fc0083b7f7e10ebed181329ca88ae/dotnet-hosting-5.0.9-win.exe

yyjdelete:
@sfoslund 
Yes, it still not work for latest version

Your OS: win10-x64(21H1, 19043.1110)
The version of the uninstall tool you're using: 1.2.206301
The full version of the hosting bundle that's failing to show up: 5.0.9
https://download.visualstudio.microsoft.com/download/pr/a0f49856-eec9-4962-8d81-b09af6be9435/1d5fc0083b7f7e10ebed181329ca88ae/dotnet-hosting-5.0.9-win.exe

sfoslund:
I was able to repro, thanks! This will be fixed in the next release.

PR #196, opened 2021-09-03
Update to support both .NET and .NET Core branding

sfoslund:
Fixes https://github.com/dotnet/cli-lab/issues/186


sfoslund:
@joeloff can I get a quick review here too?

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
```diff
@@ -72,7 +72,7 @@ internal static bool IsNetCoreBundle(string displayName, string displayVersion,
                     (displayName.IndexOf("Visual Studio", StringComparison.OrdinalIgnoreCase) < 0) &&
                     (displayName.IndexOf("VS 2015", StringComparison.OrdinalIgnoreCase) < 0) &&
                     (displayName.IndexOf("Local Feed", StringComparison.OrdinalIgnoreCase) < 0) &&
-                    ((displayName.IndexOf(".NET Core", StringComparison.OrdinalIgnoreCase) >= 0) ||
+                    ((displayName.IndexOf(".NET", StringComparison.OrdinalIgnoreCase) >= 0) ||
```
joeloff, line 75: Does this risk pickup on .NET Framework items? I don't think it should since .NET FX is part of Windows, but I have seen products call out other products like "XYZ for .NET 4.6".  Something we can maybe look into later

joeloff (approved):

PR #211, opened 2021-10-29
Adding arm64 support

sfoslund:


sfoslund:
@joeloff @ericstj I'm working on debugging this for win-arm64 and I'm confused by what the arm64 6.0 installer is writing to the registry. I'm seeing some .NET uninstall strings under `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall` but there doesn't seem to be a sdk entry, the closest one I found was `Microsoft .NET Toolset 6.0.100 (arm64)`, is that just an oddly formed SDK entry? 

I'm also trying to install the latest x64 windows installers from the installer repo grid on my test arm64 machine and it's giving me an unsupported error. Is that expected?

joeloff:
If you're installing x64 on arm64, you will need devices that support Windows 11 AFAIK.

The toolset entry is from the SDK's primary MSI

sfoslund:
@joeloff @ericstj I'm seeing some discrepancies between the x64 and arm64 registry data that gets laid out by the installers. With x64 sdk installs we always have a top level registry entry with a display name along the lines of `Microsoft .NET SDK 5.0.402 (x64)` but I'm not seeing anything analogous after installing an arm64 sdk on an arm64 machine. In fact, the uninstall tool code isn't picking up anything that it can uninstall because none of the .NET entries in the registry have a `QuietUninstallString` value and all of the `UninstallString` values begin with `MsiExec.exe`, which we specifically filter out.

Were any of these changes to how the registry keys are written intentional? After I install the SDK, it also doesn't show up in add/remove programs-- have other people tested this? Is there something wrong with my specific installation or is this a bigger problem?

ericstj:
We didn't change any of that with x64 on ARM64 and ARM64 SDK's did pre-exist before changes this release.  On my arm64 notebook I'm seeing entries in ARP.  They are created in the 32-bit registry view, here : `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall`.  I believe this happens because the SDK's bootstrapper writes the entries and it is x86.

ericstj:
Here's some screenshots:
arp_arm64
reg_uninstall_arm64

sfoslund:
Ah, okay, I've tried again with a clean machine and I'm able to see it in add/remove programs now, must have been something wrong with my set up. @ericstj  I'm seeing registry entries in both the `WOW6432Node` key and without it, do you know why that is? Which one is 'right'?

sfoslund:
The latest commit gets the tool working on windows 10 arm64 machines, still need to test on win11 with x64 emulated installs.

ericstj:
I see,  I guess they should be the same.  Would need to have a look at the bootstrapper code to be sure.  @NikolaMilosavljevic @joeloff

ericstj:
Actually, observing the registry keys on my machine I don't see any duplication.  There are many MSIs which write entries to the non-redirected registry.  Those all set `SystemComponent` = 1 which means that they don't get an entry in Add/Remove Programs UI.  The SDK bundle itself is only written once and under `Wow6432Node`.  It does not set `SystemComponent` which means it is visible in Add/Remove Programs.  It's responsible for uninstalling those other MSIs you see in the registry that are hidden from ARP.

sfoslund:
Tested on win11 arm64 with both arm64 and x64 installs and this is looking good.

joeloff:
Bundles are always 32-bit, regardless of what they install. The MSIs will go to the appropriate hive, but bundles should land under Wow6432Node.

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
```diff
@@ -62,11 +68,10 @@ private static bool IsNetCoreBundle(RegistryKey uninstallKey)
 
             return IsNetCoreBundle(uninstallKey.GetValue("DisplayName") as string,
                 uninstallKey.GetValue("DisplayVersion") as string,
-                uninstallKey.GetValue("UninstallString") as string,
-                uninstallKey.GetValue("BundleVersion") as string);
```
joeloff, line 66: Any reason for removing the BundleVersion. All the BundleXXX properties are present for bundles, but absent for MSIs, which means you could potentially pick up MSIs and classify them as bundles.

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
```diff
@@ -53,6 +35,30 @@ public virtual IEnumerable<Bundle> GetAllInstalledBundles()
...
+                    .OpenSubKey("SOFTWARE")
+                    .OpenSubKey("Microsoft")
+                    .OpenSubKey("Windows")
+                    .OpenSubKey("CurrentVersion")
+                    .OpenSubKey("Uninstall");
+
+                var names = uninstalls.GetSubKeyNames();
+
+                return names
+                    .Select(name => uninstalls.OpenSubKey(name))
```
joeloff, line 52: Performance is probably not a concern, but each registry key you open will create an open file handle.

sfoslund: Yeah it's not a big concern here. But is there a way to avoid opening these?

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
sfoslund: I think this was leftover from debugging, I'll add it back

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
joeloff: For the first part, you could try something like OpenSubKey(Path.Combine("X,","Y", "Z"). For the second part, the only option there is to pull the data you need out of the registry keys and return it using a separate class that only has the properties you care about. It's probably not a big deal given that this runs on demand and only once each time you execute the tool

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
```diff
@@ -53,6 +35,30 @@ public virtual IEnumerable<Bundle> GetAllInstalledBundles()
...
+        private IEnumerable<RegistryKey> GetNetCoreBundleKeys(RegistryKey uninstallKey)
+        {
+            try
+            {
+                var uninstalls = uninstallKey
+                    .OpenSubKey("SOFTWARE")
+                    .OpenSubKey("Microsoft")
+                    .OpenSubKey("Windows")
+                    .OpenSubKey("CurrentVersion")
+                    .OpenSubKey("Uninstall");
```
ericstj, line 47: Here you can make a single call by providing the path
```suggestion
                    .OpenSubKey(@"SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall");
```

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
ericstj: These disposable objects end up leaking here if they fail the where clause.

sfoslund: Makes sense, thanks!

src/dotnet-core-uninstall/Windows/RegistryQuery.cs
```diff
@@ -146,20 +153,35 @@ private static BundleArch GetBundleArch(string displayName, string bundleCachePa
         {
             const string x64String = "x64";
             const string x86String = "x86";
+            const string arm64String = "arm86";
```
joeloff, line 156: Should this be arm64 instead of arm86?

sfoslund: Yes, thanks!

joeloff (approved):

PR #225, opened 2022-02-25
Update publishing for official builds

joeloff:


.vsts-ci.yml
```diff
@@ -44,10 +44,15 @@ jobs:
...
+                 /p:DotNetPublishBlobFeedKey=$(dotnetfeed-storage-access-key-1)
+                 /p:DotNetPublishBlobFeedUrl=$(_PublishBlobFeedUrl)
+                 /p:DotNetPublishToBlobFeed=$(_DotNetPublishToBlobFeed)
+                 /p:DotNetSymbolServerTokenMsdl=$(microsoft-symbol-server-pat)
+                 /p:DotNetSymbolServerTokenSymWeb=$(symweb-symbol-server-pat)
                  /p:SignCoreDotnetCoreUninstall=true
                  /p:TeamName=$(_TeamName)
-                 /p:DotNetPublishToBlobFeed=$(_DotNetPublishToBlobFeed)
                  /p:OfficialBuildId=$(BUILD.BUILDNUMBER)
+                 /p:DotNetPublishUsingPipelines=$(_PublishUsingPipelines)
```
riarenas, line 55: I don't see the post-build.yml template being added anywhere, and that will be required to be able for publishing to work.

The build will also have to be transitioned to using stages so that you can add the post-build stages. 

I'd recommend pushing a branch with these changes to the internal mirror to see if the build works out and at least all the required stages show up.

riarenas: I think you'll need to follow all the steps in this https://github.com/dotnet/arcade/blob/main/Documentation/CorePackages/Publishing.md#basic-onboarding-scenario-for-new-repositories-to-the-current-publishing-version-v3

joeloff: Yeah, I figured I'd need to to something like that otherwise the testing/fixing is going to be painful.

ghost (approved):

marcpopMSFT (approved):

PR #239, opened 2022-10-04
Updated various versions to fix some transitive CG issues

marcpopMSFT:


marcpopMSFT:
@baronfel I'm sort of blindly updating all of the packages in this project as CG is flagging a transitive dependency and it's unclear which one. When I updated S.CL, I got some build failures but I couldn't find guidance on how to solve them.  Suggestions?

C:\repos\cli-lab\src\dotnet-core-uninstall\Shared\Commands\UninstallHelpBuilder.cs(14,36): error CS0246: The type or namespace name 'ICommand' could not be found (are you missing a using directive or an assembly reference?) [C:\repos\cli-lab\src\dotnet-core-uninstall\dotnet-core-uni
nstall.csproj]

baronfel:
This is going to be a whole thing - you're hitting all of the beta3 and beta4 changes at once here. The error you're seeing comes from changes to the HelpBuilder, but after you fix that all of the Option/Argument changes flow to the surface.

marcpopMSFT:
So looks like the build is still failing but this time on the Azure upgrade. I'm not sure if there is guidance on how to fix those out there somewhere.

marcpopMSFT:
@baronfel I got a bit further on this. I removed the azure blob component that was old and breaking the build as it appears it was just added here because it was copied over from installer. I also noticed that CommandLineParseResult was reporting an error because the environment command line args was including the path to the dll itself.

At this point, I'm not sure what to investigate next. I see that it returns a CommandLineConfigs object and calls invoke but I don't see what's getting invoked. I set a breakpoint in ListCommandExec.cs but it's not getting there (this is with a simple list call).  At this point, I think it's the S.CL invocation that's off.

PR #268, opened 2023-10-03
Update for 8 --> 9

Forgind:
Increases the maximum allowed version to 9


test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
```diff
@@ -27,8 +27,8 @@ public class VSVersionTests
         [InlineData(new string[] { "2.2.100", "2.2.200" }, new bool[] { false, false })]
         [InlineData(new string[] { "2.2.100", "2.2.200", "2.2.300" }, new bool[] { false, true, false })]
         [InlineData(new string[] { "3.0.0", "3.0.1", "5.0.0" }, new bool[] { true, true, false })]
-        [InlineData(new string[] { "5.0.0", "5.0.1", "6.0.1" }, new bool[] { true, false, true })]
-        [InlineData(new string[] { "6.0.0", "6.0.1", "7.0.0" }, new bool[] { true, true, false })]
+        [InlineData(new string[] { "6.0.0", "6.0.1", "7.0.1" }, new bool[] { true, false, false })]
```
Forgind, line 30: @baronfel or @joeloff,

This is the part I was least confident in as far as test changes. From what I can tell, if you provide a certain set of versions, it _always_ says the highest version is required. So how could the "allowed" list ever end in true? I made it false.

Forgind: I did note that main built, so I think there's something I'm missing here. Just wanted to call out my confusion.

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -13,33 +13,34 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.VSVersioning
...
+            { (new SemanticVersion(2, 0, 0), new SemanticVersion(2, 1, 300)), string.Format(LocalizableStrings.WindowsRequirementExplanationString, "") },
+            { (new SemanticVersion(2, 1, 300), new SemanticVersion(2, 1, 600)), string.Format(LocalizableStrings.WindowsRequirementExplanationString, " 2017") },
+            { (new SemanticVersion(2, 1, 600), new SemanticVersion(2, 1, 900)), string.Format(LocalizableStrings.WindowsRequirementExplanationString, " 2019") },
+            { (new SemanticVersion(2, 2, 100), new SemanticVersion(2, 2, 200)), string.Format(LocalizableStrings.WindowsRequirementExplanationString, " 2017") },
+            { (new SemanticVersion(2, 2, 200), new SemanticVersion(2, 2, 500)), string.Format(LocalizableStrings.WindowsRequirementExplanationString, " 2019") },
+            { (new SemanticVersion(3, 0, 100), new SemanticVersion(5, 0, 600)), string.Format(LocalizableStrings.WindowsRequirementExplanationString, " 2019") },
+            { (new SemanticVersion(5, 0, 600), new SemanticVersion(6, 0, 600)), string.Format(LocalizableStrings.WindowsRequirementExplanationString, " 2019") },
```
marcpopMSFT, line 29: lines 29 and 30 should just be the line 29 from before. Basically, this maps SDKs to VS versions. Since 2022 it everything from 6.0.100 forward, that one line doesn't need to change.

test/dotnet-core-uninstall.Tests/Shared/Commands/CommandBundleFilterTests.cs
```diff
@@ -36,9 +36,9 @@ public class CommandBundleFilterTests
         };
 
         [WindowsOnlyTheory]
-        [InlineData("remove --all --sdk", new string[] { "1.0.0", "1.0.1" })]
-        [InlineData("dry-run --all --sdk", new string[] { "1.0.0", "1.0.1" })]
-        [InlineData("whatif --all --sdk", new string[] { "1.0.0", "1.0.1" })]
+        [InlineData("remove --all --sdk", new string[] { "1.0.0", "1.0.1", "7.0.1" })]
```
marcpopMSFT, line 39: Why was 7 needed to be added here? (I'm just comparing to our last TFM update and looking at other data inputs for this test method)

test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
marcpopMSFT: We need to dig into this as this doesn't match the pattern i would expect from past history. I'd have to run the test to know for sure why it works this way.

test/dotnet-core-uninstall.Tests/Shared/Commands/CommandBundleFilterTests.cs
Forgind: To be totally honest, I have no recollection of anything having to do with this PR anymore, so I have no idea.

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
Forgind: So basically revert the meaningful changes to this file other than increasing the semantic version upper limit?

marcpopMSFT: I believe so yes. You'll still need the spelling fix obviously.

test/dotnet-core-uninstall.Tests/Shared/Commands/CommandBundleFilterTests.cs
marcpopMSFT: ahh, so in the other file we need to add additional buckets?

Forgind: If we don't want 7.0.1 to be uninstallable, then yes. But then I have to ask why 7.0.1 shouldn't be uninstallable...and also why some of those other versions shouldn't be uninstallable. Are we really going to say we can't uninstall 1.0.0 unless we have 1.1.0? Or that we want to keep both 2.1.200 and 2.1.300? Most of those are out-of-support; part of me wonders if we can just delete everything before the 3.0.100 line. This is definitely much more your area, though, so if you think we should just add more buckets, I can do that.

marcpopMSFT: My understanding is that the top entry for each band is uninstallable because it could be used by VS or VSMac. I'd have to dig into the code to confirm that but the help indicates that.

The more I dig into these tests, the more I think the tests kind of wonky. I tried to improve them but it's a lot to unwind. I'll poke more at a simpler test fix.

@joeloff @baronfel thoughts?

joeloff: Ugh, the tool is making silly guesses. we haven't shipped bundles in VS since 16.2 and .NET Core 2.2. A standalone bundle doesn't imply that VS is using it, nor does the highest band because you can have an unlimited number of instances from 15.11 through 17.9 covering each feature band.

Forgind: Sounds like joeloff recommends deprecating the uninstall tool entirely since we can't get it right anyway 😜

marcpopMSFT: @joeloff are you saying the tool is silly or the tests are silly? I think I'm close to getting a simple fix for the tests which adding in the buckets we were talking about above and adds in new items in the version object. I am just not sure about some of the true/false values for some of the other tests. I'll push something in the morning.

test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
```diff
@@ -26,10 +26,10 @@ public class VSVersionTests
         [InlineData(new string[] { "2.1.500", "2.1.400", "2.1.600" }, new bool[] { false, true, false })]
         [InlineData(new string[] { "2.2.100", "2.2.200" }, new bool[] { false, false })]
         [InlineData(new string[] { "2.2.100", "2.2.200", "2.2.300" }, new bool[] { false, true, false })]
-        [InlineData(new string[] { "3.0.0", "3.0.1", "5.0.0" }, new bool[] { true, true, false })]
```
marcpopMSFT, line 29: @Forgind @joeloff I think I figured out these tests now. The WindowsVersionDivisionsToExplanation dictionary is only used in ApplyWindowsVersionDivisions. This method is applying the buckets to the SDKs only. So this test was previously using runtime versioning which didn't fit into the buckets. To fix this, I modified the test to use SDK versions that fit within the buckets and the test passes now.

test/dotnet-core-uninstall.Tests/Shared/Commands/CommandBundleFilterTests.cs
joeloff: The tool is silly :). Back when this was designed, .NET used to ship the standalone SDK bundles in VS. Bundles in VS require overhead to track. For example, we shipped both 1.0 and 1.1 SDKs in VS 15 (eventually 2.0, 2.1, and 2.2). The tools is simply making an assumption that if you have something like the 1.1.6 SDK (which was the last version we shipped for 1.1) then it's likely that it's used by VS. And leaving the highest version of a major.minor .NET SDK on the machine is less likely to break someone.

joeloff (approved):
I'm okay with this for now to unblock getting an updated release out.

Issue #265, opened 2023-08-03
Support uninstallation of latest preview versions


pjcollins:
We have some scenarios where it could be useful to uninstall .NET previews from the latest band (currently 8.0) as we move from preview to preview, however this does not appear to be supported:

```
./dotnet-core-uninstall list  

.NET Core SDKs:
  8.0.100-preview.7.23376.3  (arm64)  [Cannot uninstall version 8.0.0 and above]   
```

I am trying to better understand this restriction. Is this not allowed because the latest SDK can have a lot of churn in between previews and the behavior of the tool may be unreliable?

Would be reasonable to add a command line param that would let users ignore this upper limit? https://github.com/dotnet/cli-lab/blob/0eab6c3a145e91f1ce16043ec63d9dfe2ac702ea/src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs#L16

baronfel:
A parameter to ignore the 'safe' limits might be interesting - really we have a roughly-yearly process of bumping this allowed upper limit ourselves. @joeloff might be time to do so again for 8.0?

ef4203:
Quick comment that this issue still persists for the latest release

jeroenhabets:
Found this issue after getting an error while trying to uninstall .Net 8 RC 1 (8.0.100-rc.1.23463.5) after the RC2 release.

mauroservienti:
It would be good to add support for .NET 8 previews. My machines start to get polluted by preview bits.

pjcollins:
I think this should be fixed for .NET 8 with https://github.com/dotnet/cli-lab/pull/268.

I also took an initial stab at allowing the upper version to be ignored for future versions in https://github.com/dotnet/cli-lab/pull/269, though this still needs some tests.

jeroenhabets:
Apologies, to nag but... Any updates on .NET 8 support?

What's holding this back? Can we help?

jrdodds:
See also issues #273 and #276

PR #269, opened 2023-10-11
[dotnet-core-uninstall] Allow preview version removal

pjcollins:
Context: https://github.com/dotnet/cli-lab/issues/265

Updates the uninstall tool to ignore the upper version check and
uninstallation restriction when the `--force` parameter is provided.

Local testing output:

```console
% sudo dotnet artifacts/bin/dotnet-core-uninstall/Debug/net8.0/dotnet-core-uninstall.dll list                   

This tool cannot uninstall versions of the runtime or SDK that are installed using zip/scripts. The versions that can be uninstalled with this tool are:

.NET Core SDKs:
  9.0.100-preview.2.24157.14  (arm64)  [Cannot uninstall version 9.0.0 and above. Use —-force to remove]             
  8.0.101                     (arm64)  [Used by Visual Studio for Mac. Specify individually or use —-force to remove]
                                                                           

.NET Core Runtimes:
  9.0.0-preview.2.24128.5  (arm64)  [Used by Visual Studio for Mac or SDKs. Specify individually or use —-force to remove]
  9.0.0-preview.2.24128.5  (x64)                                                                                          
  9.0.0-preview.2.24128.4  (arm64)                                                                                        
  9.0.0-preview.2.24128.4  (x64)  
  8.0.1   (arm64)  [Used by Visual Studio for Mac or SDKs. Specify individually or use —-force to remove]
  8.0.1   (x64)                                                                                          

% sudo dotnet artifacts/bin/dotnet-core-uninstall/Debug/net8.0/dotnet-core-uninstall.dll remove --sdk 9.0.100-preview.2.24157.14
Uninstallation not allowed. This tool cannot uninstall .NET Core SDKs with version 9.0.0 or above. Use --force to ignore this restriction.

% sudo dotnet artifacts/bin/dotnet-core-uninstall/Debug/net8.0/dotnet-core-uninstall.dll remove --sdk 9.0.100-preview.2.24157.14 --force
The following items will be removed:
  Microsoft .NET Core SDK 9.0.100-preview.2.24157.14 (x64)

To avoid breaking Visual Studio for Mac or other problems, read https://aka.ms/dotnet-core-uninstall-docs.

Do you want to continue? [y/n] y
Uninstalling: Microsoft .NET Core SDK 9.0.100-preview.2.24157.14 (x64).
```

### DIFF
--- a/src/dotnet-core-uninstall/Shared/Configs/CommandLineConfigs.cs
+++ b/src/dotnet-core-uninstall/Shared/Configs/CommandLineConfigs.cs
@@ -307,6 +307,11 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Configs
                 archSelection;
         }
 
+        public static bool ForceArgumentProvided(this ParseResult parseResult)
+        {
+            return parseResult.FindResultFor(ForceOption) != null;
+        }
+
         public static VerbosityLevel GetVerbosityLevel(this CommandResult commandResult)
         {
             var optionResult = commandResult.FindResultFor(VerbosityOption);
--- a/src/dotnet-core-uninstall/LocalizableStrings.Designer.cs
+++ b/src/dotnet-core-uninstall/LocalizableStrings.Designer.cs
@@ -596,7 +596,7 @@ namespace Microsoft.DotNet.Tools.Uninstall {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Uninstallation not allowed. This tool cannot uninstall .NET Core SDKs with version {0} or above..
+        ///   Looks up a localized string similar to Uninstallation not allowed. This tool cannot uninstall .NET Core SDKs with version {0} or above. Use --force to ignore this restriction..
         /// </summary>
         internal static string UninstallNotAllowedExceptionFormat {
             get {
@@ -641,7 +641,7 @@ namespace Microsoft.DotNet.Tools.Uninstall {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to Cannot uninstall version {0} and above.
+        ///   Looks up a localized string similar to Cannot uninstall version {0} and above. Use —-force to remove.
         /// </summary>
         internal static string UpperLimitRequirement {
             get {
--- a/src/dotnet-core-uninstall/LocalizableStrings.resx
+++ b/src/dotnet-core-uninstall/LocalizableStrings.resx
@@ -298,7 +298,7 @@ Do you want to continue? [y/n] </value>
     <value>Remove the specified .NET Core SDKs or Runtimes.</value>
   </data>
   <data name="UninstallNotAllowedExceptionFormat" xml:space="preserve">
-    <value>Uninstallation not allowed. This tool cannot uninstall .NET Core SDKs with version {0} or above.</value>
+    <value>Uninstallation not allowed. This tool cannot uninstall .NET Core SDKs with version {0} or above. Use --force to ignore this restriction.</value>
   </data>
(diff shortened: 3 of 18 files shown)


src/dotnet-core-uninstall/Shared/Configs/CommandLineConfigs.cs
```diff
@@ -133,6 +133,10 @@ internal static class CommandLineConfigs
             RuntimeInfo.RunningOnWindows ? LocalizableStrings.ForceOptionDescriptionWindows
             : LocalizableStrings.ForceOptionDescriptionMac);
 
+        public static readonly Option IgnoreUpperLimitOption = new Option(
+            "--ignore-upper-limit",
```
pjcollins, line 137: I'm not sure if it makes more sense to add this new `--ignore-upper-limit` param, or if we should just extend `--force` to allow uninstallation of versions larger than the declared upper limit.

pjcollins: @joeloff @baronfel I was looking at improving preview uninstall support for future versions. My initial thought was to add this `--ignore-upper-limit` parameter, but after thinking a bit more I think the UX may be better if we overload `--force` to also handle this case instead?

joeloff:
cc @Forgind

marcpopMSFT:
@baronfel per our chat earlier about removing the upper limit. This PR provides an alternative option for customers though is not exacty what we were talking about.

pjcollins:
I reworked this to use the `--force` argument to uninstall versions over the upper limit, rather than adding a new parameter.

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -103,7 +104,9 @@ public static IEnumerable<Bundle> GetUninstallableBundles(IEnumerable<Bundle> bu
         {
             var (bundlesByDivisions, remainingBundles) = ApplyVersionDivisions(allBundles);
 
-            var bundlesAboveUpperLimit = remainingBundles.Where(bundle => bundle.Version.SemVer >= UpperLimit);
+            var bundlesAboveUpperLimit = CommandLineConfigs.CommandLineParseResult.ForceArgumentProvided()
```
Forgind, line 107: I don't think this does what you might expect? I think this just changes a string, which shouldn't really matter for uninstallation purposes. If the user specifies force (or however we decide they should be able to request uninstalling versions above the upper limit), I think we should instead add an extra version division [UpperLimit, infinity) and treat it the same as the rest.

pjcollins: Thanks, this was left over from some of my initial changes and I forgot to back it out.

With respect to the extra version division suggestion, are you suggesting that we remove the "force argument provided" check around [CommandBundleFilter.FilterRequiredBundles](https://github.com/dotnet/cli-lab/pull/269/files#diff-bbb2c6b491062b0828246e408aee2723c174f1aa44318314f96886b47aec406dR71) and instead check if this param is provided and update the version groups created in VisualStudioSafeVersionsExtractor.ApplyWindowsVersionDivisions / VisualStudioSafeVersionsExtractor.ApplyMacVersionDivisions accordingly?

Forgind: Yes; that's what I was thinking. I haven't spent that much time with cli-lab, though, so it would probably be good to get signoff from someone else who's commented on this PR.

PR #286, opened 2024-03-11
Split yml

Forgind:
Splits our yml to support using the new templates.

This will require redirecting the yml we use for PR builds as soon as this is merged.


.vsts-ci.yml
```diff
@@ -15,154 +16,163 @@ pr:
...
+  - repository: 1ESPipelineTemplates
+    type: git
+    name: 1ESPipelineTemplates/1ESPipelineTemplates
+    ref: refs/tags/release
+extends:
+  template: v1/1ES.Official.PipelineTemplate.yml@1ESPipelineTemplates
+  parameters:
+    pool:
+      name: $(DncEngInternalBuildPool)
+      image: 1es-windows-2022-pt
```
marcpopMSFT, line 30: dnceng says the non -pt images have now been updated to support the template migration. You can switch back now or later (they said the -pt ones will go away eventually). Either way.

.vsts-ci.yml
```diff
@@ -15,154 +16,163 @@ pr:
...
+              ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+                name: $(DncEngInternalBuildPool)
+                image: 1es-windows-2022
+                os: windows
+            variables:
+            - name: _RID
+              value: win-x86
+            - name: _BuildConfig
+              value: Release
+            - ${{ if or(eq(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'PullRequest')) }}:
```
marcpopMSFT, line 65: All the public and PR code can probably be removed later as well since we're going into a split pipeline mode.

marcpopMSFT (approved):

.vsts-ci.yml
Forgind: MiYanni suggested yesterday that we should keep the split pipelines in the hope that we'd eventually be merging them with the -pr version again anyway.

MiYanni: I intentionally left all public/internal logic in my PR too. I would not recommend removing it as there is a likelihood of wanting it if we need to do 1ES templates for public pipelines too in the future.

marcpopMSFT: ahh, makes sense. I think he's optimistic but it is safer to leave this in place for now just in case.

PR #290, opened 2024-03-12
Rip out the static ParseResult and instead compute a PR once and flow it through the command handlers

### LINKED ISSUE #287: [NETSDKE2E][MAC] Throw an unhandled exception in dotnet-core-uninstall list output.
### To Reproduce
1.install dotnet-core-uninstall(1.7.0) from https://dnceng.visualstudio.com/_apis/resources/Containers/16714346/drop-osx?itemPath=drop-osx%2Fdotnet-core-uninstall.tar.gz
2. Use the following commands to install the tool:
```
mkdir -p ~/dotnet-core-uninstall
tar -zxf dotnet-core-uninstall.tar.gz -C ~/dotnet-core-uninstall
cd ~/dotnet-core-uninstall
./dotnet-core-uninstall list
```
3.Install SDK 8.0.201from https://dotnet.microsoft.com/en-us/download/dotnet/8.0
4. Run "./dotnet-core-uninstall list"
5.check the output of dotnet-core-uninstall list

### Expected Result:
The version(s) of the .NET 8 SDK you installed appear in the output.
### Actual Result:
Throw an unhandled exception in dotnet-core-uninstall list output.
image

### dotnet --info:
image

### Note:
This issue also repro on.NET 6/7.

baronfel:
Fixes #287 

This removes the use of [`Environment.GetCommandLineArgs()` in the static constructor]([`remove-static-argv-access?expand=1`#diff-2ae2e07177](https://github.com/dotnet/cli-lab/compare/remove-static-argv-access?expand=1#diff-2ae2e07177a9c78d9195f6771080cbb64efb6ea45a6c9e814c8022d2c3f13618)L244) of the `CommandLineConfigs` type. Since this was being used in other commands directly, we now have to inject a ParseResult into the various commands. This also means that the current implementation of the `--version` option didn't work anymore, so I did what is the common/conventionally-done thing which is pulling `--version` out into a top-level command.


src/dotnet-core-uninstall/Shared/Commands/CommandBundleFilter.cs
```diff
@@ -79,24 +59,15 @@ public static IEnumerable<Bundle> GetFilteredBundles(IEnumerable<Bundle> allBund
...
         }
 
-        public static void HandleVersionOption()
-        {
-            if (CommandLineConfigs.CommandLineParseResult.FindResultFor(CommandLineConfigs.VersionOption) != null)
-            {
-                Console.WriteLine(_assemblyVersion.Value);
-                Environment.Exit(0);
-            }
-        }
```
baronfel, line 98: this becomes a top-level option in the CLI parser

joeloff (approved):

test/dotnet-core-uninstall.Tests/Shared/IntegrationTests/IntegrationTests.cs
```diff
@@ -1,21 +0,0 @@
...
-        [Fact]
-        internal async void VersionCommandOutputsVersionInfo()
-        {
-            var console = new TestConsole();
-            var output = await CommandLineConfigs.UninstallCommandParser.InvokeAsync(new string[1] { "--version" }, console);
-            output.Should().Be(0);
-            console.Out.ToString().Should().Match("*.*.*");
-        }
-    }
-}
```
baronfel, line 21: I removed this test because it's not useful/relevant - the S.CL framework provides `--version` and we should just rely on it.

PR #307, opened 2024-09-12
Enable default SDL checks on build

marcpopMSFT:


.config/tsaoptions.json
```diff
@@ -0,0 +1,10 @@
+{
+    "instanceUrl": "https://devdiv.visualstudio.com/",
+    "template": "TFSDEVDIV",
+    "projectName": "DEVDIV",
+    "areaPath": "DevDiv\\NET Tools\\SDK",
+    "iterationPath": "DevDiv",
+    "notificationAliases": [ "dotnetdevexcli@microsoft.com" ],
+    "repositoryName":"cli-lab",
```
MiYanni, line 8:
```suggestion
    "repositoryName": "cli-lab",
```

.vsts-ci.yml
```diff
@@ -46,6 +51,11 @@ extends:
           timeoutInMinutes: 180 # increase timeout since BAR publishing might wait a long time
           jobs:
           - job: Windows
+            templateContext:
+              outputs:
+                - output: pipelineArtifact
+                  targetPath: $(Build.SourcesDirectory)/artifacts/bin/dotnet-core-uninstall
+                  artifactName: UninstallTool
```
MiYanni, line 58: The artifacts for Windows are already published here: https://github.com/dotnet/cli-lab/pull/307/files#diff-fb7d8d37fd9e28ca688fd102422af3e1d1df13a50d7f2deb2e145aa5ef449057R118-R124

OSX has a similar step to publish their artifacts. I understand from the thread that you wanted it to recognize the outputs for the SDL tools, but double publishing isn't the right way to approach the problem. Which tool didn't recognize the outputs properly? I thought it was binskim, but I don't see that in your `sdl` node.

marcpopMSFT: I didn't even realize that this would do a second publish as that's not obvious at all from either the examples or documentation. If this is not the way to override the output location then then I'm not sure how to do it. There docs have patter matching but don't have a way to override the output. I assumed this was just giving context to the run to know where the binaries were:
https://dev.azure.com/1ESPipelineTemplates-PPE/1ESPipelinesTest/_git/1ESPipelinesTest-SDLAnalysis?path=/Samples/BinaryAnalysis/Windows/Binskim-WithTargetFiles.yml&version=GC800b9ae2ff46a159883e6f4a5e9fc54ae86a3b72

Seems like I have to ping 1es then as this isn't the right solution.

marcpopMSFT: Found a way to do this using the analyzeTargetGlob parameter.

MiYanni (approved):

PR #313, opened 2024-10-24
Create a macOS universal binary

baronfel:
Adds a target that
* creates arch-specific macos binaries
* stitches them together using `lipo` to create a universal binary (thanks @rolfbjarne!)

Things we need to think about:
* how are these signed/notarized?
* how is the universal binary passed to the rest of the build process in the azdo pipelines?
* How does the azdo pipeline on mac call the build? Because we'd need it to either call this new target directly or have this target be a dependency of the existing build entry point

cc @joeloff / @marcpopMSFT


src/dotnet-core-uninstall/dotnet-core-uninstall.csproj
```diff
@@ -37,4 +37,20 @@
...
+    </ItemGroup>
+
+    <MSBuild Projects="@(_RidSpecificBuild)" BuildInParallel="true" Targets="Publish;PublishItemsOutputGroup">
+     <Output TaskParameter="TargetOutputs" ItemName="_RidSpecificSingleFile" />
+    </MSBuild>
+    <PropertyGroup>
+      <_LipoOuputPath>$(PublishDir)\osx-universal\$(AssemblyName)</_LipoOuputPath>
+    </PropertyGroup>
+
+    <Exec Command="lipo @(_RidSpecificSingleFile, ' ') -create -output $(_LipoOuputPath)" Condition="$([MSBuild]::IsOSPlatform('OSX'))" />
```
rolfbjarne, line 53: I don't know if you care, but this won't work if any of these paths have spaces in them.

src/dotnet-core-uninstall/dotnet-core-uninstall.csproj
```diff
@@ -37,4 +37,20 @@
...
+    <ItemGroup>
+      <_Rids Include="osx-x64;osx-arm64" />
+      <_RidSpecificBuild Include="$(MSBuildThisFile)" AdditionalProperties="RuntimeIdentifier=%(_Rids.Identity);PublishSingleFile=true;Configuration=Release" />
+    </ItemGroup>
+
+    <MSBuild Projects="@(_RidSpecificBuild)" BuildInParallel="true" Targets="Publish;PublishItemsOutputGroup">
+     <Output TaskParameter="TargetOutputs" ItemName="_RidSpecificSingleFile" />
+    </MSBuild>
+    <PropertyGroup>
+      <_LipoOuputPath>$(PublishDir)\osx-universal\$(AssemblyName)</_LipoOuputPath>
```
rolfbjarne, line 50:
```suggestion
      <_LipoOutputPath>$(PublishDir)\osx-universal\$(AssemblyName)</_LipoOutputPath>
```

src/dotnet-core-uninstall/dotnet-core-uninstall.csproj
rolfbjarne, line 53:
```suggestion
    <Exec Command="lipo @(_RidSpecificSingleFile, ' ') -create -output $(_LipoOutputPath)" Condition="$([MSBuild]::IsOSPlatform('OSX'))" />
```

baronfel:
Tested this on a m2 machine and it works as expected:

```
% file dotnet-core-uninstall 
dotnet-core-uninstall: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64:Mach-O 64-bit executable arm64]
dotnet-core-uninstall (for architecture x86_64):	Mach-O 64-bit executable x86_64
dotnet-core-uninstall (for architecture arm64):	Mach-O 64-bit executable arm64
```

joeloff:
Don't think we do notarization for this.

src/dotnet-core-uninstall/dotnet-core-uninstall.csproj
```diff
@@ -37,4 +37,23 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <Target Name="PublishMacOSUniversalBinary">
```
joeloff, line 40: who/what calls this target?

joeloff: probably here: https://github.com/dotnet/cli-lab/blob/25e91f1d39ad446131c0fe4bc998c5a33e766773/.vsts-ci.yml#L137C1-L144C33

You can tweak the .yml, then push your branch internal to trigger an official build against your branch, that's the only way to verify yaml changes

marcpopMSFT:
@baronfel is this PR still something you think we'd want? I pushed main into it just to get it off of a scan radar for some old files that needed updating.

baronfel:
IMO no - macOS is pretty exclusively arm64 by this point.

Issue #311, opened 2024-10-18
[NETSDKE2E][MAC]The text "Used by VS for Mac..." next to net 9 SDK needs to be updated as VS Mac has been end of life.


vdanche:
To Reproduce
1.Prepare a clean Intel
2.install dotnet-core-uninstall(1.7.550802)
3. Use the following commands to install the tool:
```
mkdir -p ~/dotnet-core-uninstall
tar -zxvf dotnet-core-uninstall.tar.gz -C ~/dotnet-core-uninstall
cd ~/dotnet-core-uninstall
./dotnet-core-uninstall list
```
4.Install 9.0.rc2 sdk and some others sdk . e.g 8.0.403 sdk...
5. Run "./dotnet-core-uninstall list"
6.check the output of dotnet-core-uninstall list

**Expected Result:**
The output of dotnet-core-uninstall list display well.

**Actual** Result:
The text "Used by VS for Mac..." next to net 9 SDK needs to be updated as VS Mac has been end of life.
![Image](https://github.com/user-attachments/assets/ceb60d6d-1ba6-4b0e-b4d0-74784385c720)

dotnet --info:
![Image](https://github.com/user-attachments/assets/d60fcf6a-a7a6-4ef0-9a26-a5c41f7147ec)

marcpopMSFT:
I kind of think we should just remove this check entirely. @baronfel @joeloff https://github.com/dotnet/cli-lab/blob/25e91f1d39ad446131c0fe4bc998c5a33e766773/src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs#L53

not a blocker.

edvilme:
> I kind of think we should just remove this check entirely. [@baronfel](https://github.com/baronfel) [@joeloff](https://github.com/joeloff)
> 
> [cli-lab/src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs](https://github.com/dotnet/cli-lab/blob/25e91f1d39ad446131c0fe4bc998c5a33e766773/src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs#L53)
> 
> Line 53 in [25e91f1](/dotnet/cli-lab/commit/25e91f1d39ad446131c0fe4bc998c5a33e766773)
> 
[...]

PR #318, opened 2025-01-17
Remove Visual Studio macOS checks

edvilme:
Addresses #311 
Visual Studio for Mac entered End Of Life support on 9.0.000. Hence, it should not show the text for VS in future versions. 

Default behavior is kept when using the `--preserve-mac-vs-sdks` flag


src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -58,18 +61,21 @@ private static (IDictionary<IEnumerable<Bundle>, string>, IEnumerable<Bundle>) A
...
-                .Select(pair => (pair as IEnumerable<Bundle>, LocalizableStrings.MacRuntimeRequirementExplanationString))
-                .ToDictionary(key => key.Item1, value => value.Item2); 
+                .ToDictionary(
+                    group => group as IEnumerable<Bundle>,
+                    group => group.Any(bundle => bundle.Version.SemVer <= MacOSVisualStudioUpperLimit) ? LocalizableStrings.MacRuntimeRequirementExplanationString : string.Empty
+                );
 
             var sdks = bundleList.Where(bundle => bundle.Version is SdkVersion);
-            if (sdks != null && sdks.Count() > 0)
+            if (sdks.Any())
```
Forgind, line 70: If sdks is null, this will throw

Forgind (changes requested):
This doesn't look right to me. It looks like you're removing the message that we need to keep it _because_ of VS For Mac (if the SDKs are of a high-enough version) but still make it an implicit requirement for uninstalling and even extends that requirement to Windows, which is not desirable. We also don't really care about VS for Mac anymore.

src/dotnet-core-uninstall/LocalizableStrings.resx
```diff
@@ -390,4 +387,10 @@ Warning: {0}: {1}
 Uninstalling this item will cause Visual Studio for to break.
 </value>
   </data>
-</root>
+  <data name="MacOSPreserveVSSdksOptionDescription" xml:space="preserve">
+    <value>Keep SDKs required for Visual Studio for Mac (no longer supported)</value>
```
edvilme, line 391: I think wording here could be improved, maybe 
```suggestion
    <value>Keep SDKs required for Visual Studio for Mac (Visual Studio for Mac is no longer supported)</value>
``` 
??

src/dotnet-core-uninstall/Shared/Configs/CommandLineConfigs.cs
```diff
@@ -135,6 +135,10 @@ internal static class CommandLineConfigs
             RuntimeInfo.RunningOnWindows ? LocalizableStrings.ForceOptionDescriptionWindows
             : LocalizableStrings.ForceOptionDescriptionMac);
 
+        public static readonly Option MacOSPreserveVSSdksOption = new Option(
+            "--preserve-mac-vs-sdks",
```
edvilme, line 139: I believe adding `mac` in the flags' name is important as to avoid confusion with Windows users (flag is only shown to macOS users, but still)

test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
```diff
@@ -147,7 +148,7 @@ private void CheckAllowed(IEnumerable<Bundle> allBundles, IEnumerable<Bundle> un
                 {
                     uninstallableBundles.Should().Contain(sdkBundles[i]);
                 }
-                else
+                else if (!RuntimeInfo.RunningOnOSX)
```
edvilme, line 151: In the new behavior, all SDK versions should be uninstallable on Mac

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -61,48 +61,52 @@ private static (IDictionary<IEnumerable<Bundle>, string>, IEnumerable<Bundle>) A
...
-                required.Add(band.Max());
+                protectedBundles.Add(band.Max());
             }
 
-            required = required.Concat(remainingBundles.Where(bundle => bundle.Version.SemVer >= UpperLimit)).ToList();
+            protectedBundles = protectedBundles.Concat(remainingBundles.Where(bundle => bundle.Version.SemVer >= UpperLimit)).ToList();
 
-            return bundles.Where(b => !required.Contains(b));
+            // Return all bundles that are not protectedBundles
+            return bundles.Except(protectedBundles);
```
Forgind, line 104: I think this is correct, but I got burned at one point making a change like this because this looks equivalent but isn't...the order doesn't matter, and you don't need duplicates, right?

test/dotnet-core-uninstall.Tests/Shared/Commands/ListCommandExecTests.cs
```diff
@@ -37,7 +37,7 @@ internal void ListCommandFilteringIsCorrectOnWindows(string bundleType, string o
         }
 
         [MacOsOnlyTheory]
-        [InlineData("sdk", "", new string[] { "3.1.0" }, new string[] { "3.0.0", "3.0.0-preview", "1.0.0", "3.0.1", "3.0.2", "3.0.2-preview1", "3.0.2-preview2", "2.1.1", "1.0.1" })]
+        [InlineData("sdk", "", new string[] {}, new string[] { "3.0.0", "3.0.0-preview", "1.0.0", "3.0.1", "3.0.2", "3.0.2-preview1", "3.0.2-preview2", "2.1.1", "1.0.1", "3.1.0" })]
```
edvilme, line 40: `"3.1.0"` included in `expectedUninstallable` as all sdks should be uninstallable on Mac

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
```diff
@@ -61,48 +61,52 @@ private static (IDictionary<IEnumerable<Bundle>, string>, IEnumerable<Bundle>) A
...
+        private static (IDictionary<IEnumerable<Bundle>, string>, IEnumerable<Bundle>) ApplyVersionDivisions(IEnumerable<Bundle> bundles, bool macOSPreserveVSSdks = false)
         {
             if (RuntimeInfo.RunningOnWindows)
             {
                 return ApplyWindowsVersionDivisions(bundles);
             }
             else
             {
-                return ApplyMacVersionDivisions(bundles);
+                return ApplyMacVersionDivisions(bundles, macOSPreserveVSSdks);
```
Forgind, line 87: Is the first divided bundles important? I'd expected to just condition having this call at all on macOSPreserveVSSdks

src/dotnet-core-uninstall/LocalizableStrings.resx
```diff
@@ -339,14 +339,11 @@ The versions that can be uninstalled with this tool are:
 </value>
   </data>
   <data name="UninstallNoOptionDescriptionMac" xml:space="preserve">
-    <value>Remove specified .NET Core SDKs or Runtimes. By default, this tool does not uninstall versions that might be needed for Visual Studio for Mac or SDKs. Read the documentation for the .NET Core Uninstall Tool at https://aka.ms/dotnet-core-uninstall-docs.</value>
+    <value>Remove specified .NET Core SDKs or Runtimes. By default, this tool does not uninstall versions that might be needed for SDKs. Read the documentation for the .NET Core Uninstall Tool at https://aka.ms/dotnet-core-uninstall-docs.</value>
```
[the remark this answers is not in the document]
edvilme: I think it referred to not uninstalling runtime versions that might be needed for VS, and hence by sdks

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
edvilme: Yes. From what I understand, while the sdks are all uninstallable now, runtimes should still be checked to avoid uninstalling a runtime associated to an sdk

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
edvilme: Yes. But to be safe I can revert

src/dotnet-core-uninstall/Shared/VSVersioning/VisualStudioSafeVersionsExtractor.cs
Forgind: If you uninstall an SDK, why do you need the runtime associated with that SDK?

edvilme: I thought it more like the other way around, when people attempt to uninstall a runtime associated with an sdk

edvilme: So then you mean it should not compare runtimes and sdks when uninstalling either?

Forgind: So like let's say you have SDKs sa, sb, sc, and sd installed right now, and you have runtime ra, rb, and rc. sa depends on ra, sb on rb, and both sc and sd depend on rc. If you're planning to uninstall sa and sc and keep sb and sd, then you should uninstall ra because no SDK that isn't being uninstalled requires it. You can't uninstall rb because you aren't uninstalling sb, and you can't install rc because sd requires it even though you're uninstalling sc. Is that clearer?

edvilme: Yes, thanks. So in essence, when uninstalling an sdk, it should also remove unused runtimes, right?

Forgind: Yeah, for mac I think that's right.

Windows get a little more confusing because we don't really look for VS instances properly, so we (conservatively) didn't uninstall runtimes VS might need. We used to do the same for mac, but since VS for mac isn't really a thing anymore, we don't have to.

Forgind (approved):
Thanks for your work on this! I think it does what I'd expect now 🙂

PR #320, opened 2025-01-23
macOS: Show correct arm64 architecture

edvilme:
Closes #297 #221 
When listing runtimes and sdk's on arm Macs, the architecture should be displayed as arm64 instead of x64.

Currently this is not figured out on a by version lookup, but based on the first directory on the sdk installs directory

`list`
image

`dry-run`
image

PR #321, opened 2025-01-27
Identify macOS runtimes correctly

edvilme:
On macOS, architecture was only checked for SDK versions, not for runtime versions. This led to incorrect architecture when listing the versions.

On runtimes, the architecture should be checked from the file `Microsoft.NETCore.App.deps.json` inside the runtime's directory. 

## Current behavior
image

## Expected behavior
image

Relates to changes in #320

### DIFF
--- a/src/dotnet-core-uninstall/MacOs/FileSystemExplorer.cs
+++ b/src/dotnet-core-uninstall/MacOs/FileSystemExplorer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text.Json;
 using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo;
 using Microsoft.DotNet.Tools.Uninstall.Shared.BundleInfo.Versioning;
 using Microsoft.DotNet.Tools.Uninstall.Shared.Configs;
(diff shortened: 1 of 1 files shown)


src/dotnet-core-uninstall/MacOs/FileSystemExplorer.cs
```diff
@@ -46,12 +47,22 @@ public virtual IEnumerable<Bundle> GetAllInstalledBundles()
...
+        private static bool IsMacx64Installation(string bundlePath, bool isSdk)
         {
             try
             {
-                var rids = File.ReadAllText(Path.Combine(sdkVersionPath, "NETCoreSdkRuntimeIdentifierChain.txt"));
-                return !rids.Contains("osx-arm64");
+                if (isSdk)
+                {
+                    var rids = File.ReadAllText(Path.Combine(bundlePath, "NETCoreSdkRuntimeIdentifierChain.txt"));
+                    return !rids.Contains("osx-arm64");
```
edvilme, line 57:
```suggestion
                    return rids.Contains("osx-x64");
```

Perhaps readability would be better if we avoid the double negative??

src/dotnet-core-uninstall/MacOs/FileSystemExplorer.cs
```diff
@@ -46,12 +47,22 @@ public virtual IEnumerable<Bundle> GetAllInstalledBundles()
...
+                }
+                else
+                {
+                    var depsJsonFile = Directory.EnumerateFiles(bundlePath, "Microsoft.*.deps.json", SearchOption.TopDirectoryOnly).FirstOrDefault();
```
edvilme, line 61: ASP.net also does have its own .deps.json file

Forgind (approved):
This seems reasonable, I think, but I don't know very much about .deps.json files. I looked around briefly and didn't see anything better, so 🙂

PR #325, opened 2025-01-31
Refactor macOS build pipeline

edvilme:
Simplifies the workflow of macOS builds to make it easier to extend afterwards (e.g., adding signing, notarization, etc)

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -16,6 +16,13 @@ pr:
     include:
     - '*'
 
+parameters:
+- name: MacOSArchitectures
+  type: object
+  default:
+  - arm64
+  - x64
+
 resources:
   repositories:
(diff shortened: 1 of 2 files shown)


Forgind:
There are parts of this (like missing hyphens) that seem like they shouldn't work to me, but if it works, I guess it works? /cc: @MiYanni for a look if you have time.

edvilme:
> There are parts of this (like missing hyphens) that seem like they shouldn't work to me, but if it works, I guess it works? /cc: @MiYanni for a look if you have time.

I think your're right, and that syntax was incorrect for `.vsts-ci.yml` which doesn't run on the PR. Fixed it now :)

MiYanni (approved):
I didn't review the original PR that added these jobs, but this solution to use a strategy matrix looks much cleaner for this repo's pipeline.

PR #328, opened 2025-02-10
Fix arcade signing

edvilme:
After updating to arcade, builds were not generating correctly.
This adds an automated PR for 1ES compliance, and removes signing targets as these are now generated automatically on Arcade

Trying to keep [8c89301](https://github.com/dotnet/cli-lab/pull/328/commits/8c89301ebab09d8aa949efc3043b704df973124a) as its own commit as to preserve git history


src/redist/redist.csproj
```diff
@@ -10,16 +10,13 @@
...
   </ItemGroup>
-
+  
   <Import Condition="'$(RID)' == 'win-x86'" Project="targets\Versions.targets" />
   <Import Condition="'$(RID)' == 'win-x86'" Project="targets\BuildCoreSdkTasks.targets" />
   <Import Condition="'$(RID)' == 'win-x86'" Project="targets\GetRuntimeInformation.targets" />
   <Import Project="targets\GenerateLayout.targets" />
   <Import Condition="'$(RID)' == 'win-x86'" Project="targets\FileExtensions.targets" />
   <Import Condition="'$(RID)' == 'win-x86'" Project="targets\GenerateMSIs.targets" />
-  <Import Condition="'$(RID)' == 'win-x86'" Project="targets\Signing.targets" />
```
edvilme, line 20: Signing.targets no longer needed as handled by Arcade now

src/redist/targets/GenerateLayout.targets
```diff
@@ -5,7 +5,6 @@
     <_PublishSingleFileForRid>true</_PublishSingleFileForRid>
     <_PublishSingleFileForRid Condition="'$(RID)' == 'win-x86'">false</_PublishSingleFileForRid>
     <_GenerateLayoutDependsOnTargets>CleanLayoutPath;PublishDotnetCoreUninstallProject</_GenerateLayoutDependsOnTargets>
-    <_GenerateLayoutDependsOnTargets Condition="'$(RID)' == 'win-x86'">CleanLayoutPath;PublishDotnetCoreUninstallProject;SignLayout</_GenerateLayoutDependsOnTargets>
```
edvilme, line 8: `SignLayout` is also now handled by Arcade now

Forgind: Why did we lose the condition?

edvilme: Because it is identical to the previous line, but adds the `SignLayout` target (which is now unused)

Forgind: Good point

Forgind (approved):

PR #329, opened 2025-02-21
Fix Windows Signing

edvilme:
MSI contents were not correctly signed on Windows. Using [`CreateLightCommandPackageDrop`](https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.Build.Tasks.Installers#task-usage) to handle re-packaging and signing of resulting .msi file.


joeloff (approved):
Verified the output with @edvilme after he ran signcheck on the MSI

Issue #304, opened 2024-05-22
The current user does not have adequate privileges


ArchieCoder:
Hi,

I'm no Mac expert. I don't know how to run the tool. The documentation does not explain what is required.

I have only one user on my Mac. My user has admin privilege. In settings, the toggle is on for the admin to allow administer my computer.

Thank you

ScottAtRedHawk:
Same issue on macOS 14.5 Sonoma.

edvilme:
Hello @ArchieCoder @ScottAtRedHawk can you please provide additional context to the error that is appearing on your computers?

You can either try building the tool with `-p:RID=osx-arm64` for Apple Silicon (M-Series Macs) or `-p:RID=osx-x64` for Intel Macs, and running the output binary in `artifacts/bin/dotnet-core-uninstall`; or use the .tar.gz in the releases section of this repo. 

Please let me know if this is helpful.

ArchieCoder:
I downloaded the tar file. Here is the output (see last line)

<img width="567" alt="Image" src="https://github.com/user-attachments/assets/43fe7379-10fa-465b-94d3-2cef40e2f0e6" />

edvilme:
Have you tried running the command with `sudo`? 
The tool attempts to uninstall from secured locations which require special permissions to remove files

ArchieCoder:
@edvilme It works with sudo thanks. Maybe this command should be in the instructions.

PR #331, opened 2025-02-27
Small refactorings

edvilme:
~Addresses #304~
Small refactorings to `IsAdmin` and other places in the execute command.

### DIFF
--- a/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
+++ b/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
@@ -155,8 +155,8 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
             {
                 return new ProcessStartInfo
                 {
-                    FileName = "rm",
-                    Arguments = $"-rf {command}",
+                    FileName = "sudo",
+                    Arguments = $"rm -rf {command}",
                     UseShellExecute = true
                 };
             }
--- a/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
+++ b/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
@@ -33,32 +33,16 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
 
         public static void Execute(IBundleCollector bundleCollector, ParseResult parseResult)
         {
+            if (!IsAdmin())
+            {
+                throw new NotAdminException();
+            }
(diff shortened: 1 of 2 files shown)


src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
```diff
@@ -33,33 +33,17 @@ private static extern IntPtr CommandLineToArgvW(
 
         public static void Execute(IBundleCollector bundleCollector, ParseResult parseResult)
         {
+            if (!IsAdmin())
```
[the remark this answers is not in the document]
edvilme: Yes. However the logic for the dry run command exists in the [`DryRunCommandExec` class](https://github.com/dotnet/cli-lab/blob/main/src/dotnet-core-uninstall/Shared/Commands/DryRunCommandExec.cs)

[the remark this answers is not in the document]
joeloff: Sounds good

src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
```diff
@@ -241,17 +215,18 @@ public static bool AskWithWarningsForRequiredBundles(IDictionary<Bundle, string>
                 Console.Write(string.Format(RuntimeInfo.RunningOnWindows ? LocalizableStrings.WindowsRequiredBundleConfirmationPromptOutputFormat :
                     LocalizableStrings.MacRequiredBundleConfirmationPromptOutputFormat, pair.Key.DisplayName, pair.Value));
                 Console.ResetColor();
-                var response = userResponse == null ? Console.ReadLine().Trim().ToUpper() : userResponse.ToUpper();
+                
+                var response = userResponse?.ToUpper() ?? Console.ReadLine().Trim().ToUpper();
+
                 if (response.Equals("N"))
```
joeloff, line 221: Should we be checking for `NO` as well, given that we check for `y` or `yes`? If we really want to restrict an interactive response, then I wonder if a loop with ReadKey is better and we only accept N/n/y/Y

joeloff (approved):

src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
joeloff: Even better.

edvilme: Updated, let me know what you think :)

PR #337, opened 2025-03-11
Update to AwesomeAssertions

marcpopMSFT:
With FluentAssertions version 8 changing their license, we're exploring alternative solutions. While we use the 6 version today, if there were a security reason to update in the future, we want to ensure we're on a supported version.

Update the addreportable call

### DIFF
--- a/test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
+++ b/test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
@@ -1,7 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Reflection.Metadata;
 using FluentAssertions;
 using FluentAssertions.Execution;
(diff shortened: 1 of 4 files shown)


test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection.Metadata;
```
danmoseley, line 4: ?

danmoseley (approved):

test/dotnet-core-uninstall.Tests/Shared/VSVersioning/VSVersionTests.cs
marcpopMSFT: I think copilot must have imagined something when fixing this and I didn't notice.

edvilme (approved):

PR #340, opened 2025-03-14
Require enter on user input

### LINKED ISSUE #339: [NETE2ESDK]It should need to press enter to confirm when type y/n to continue with command uninstall
Build: dotnet-core-uninstall-1.7.615501
OS: Win 11 24h2 X64

Repro steps:
1. Run `dotnet-core-uninstall remove --sdk {version}` or  `dotnet-core-uninstall remove --runtime {version}`
2. Typing y or n with prompts 'Do you want to continue? 
3. Observe

Expected Result:
It needs to press enter to confirm before continuing.  Take the behavior in v1.7.550802 for example
Image

Actually Result:
Typing y or n without confirming will continue.
Image

edvilme:
Fixes #339 by requiring the user to press enter to confirm when typing 'y' or 'n' to continue with the command uninstall.

### DIFF
--- a/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
+++ b/src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
@@ -193,7 +193,7 @@ namespace Microsoft.DotNet.Tools.Uninstall.Shared.Commands
             Console.Write(string.Format(RuntimeInfo.RunningOnWindows ? LocalizableStrings.WindowsConfirmationPromptOutputFormat :
                 LocalizableStrings.MacConfirmationPromptOutputFormat, displayNames));
 
-            var response = userResponse?.ToUpper() ?? Console.ReadKey().KeyChar.ToString().ToUpper();
+            var response = (userResponse ?? Console.ReadLine()).ToUpper();
 
             if (response.Equals("Y"))
             {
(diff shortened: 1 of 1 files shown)


src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
```diff
@@ -216,13 +216,13 @@ public static bool AskWithWarningsForRequiredBundles(IDictionary<Bundle, string>
...
                 
-                var response = userResponse?.ToUpper() ?? Console.ReadKey().KeyChar.ToString().ToUpper();
+                var response = (userResponse ?? Console.ReadLine()).ToUpper();
 
                 if (response.Equals("N"))
                 {
                     return false;
                 }
-                if (!(response.Equals("Y") || response.Equals("YES")))
+                if (!response.Equals("Y"))
```
MiYanni, line 225: So, `yes` doesn't work anymore? I guess it only asks for `y/n`, so that's fine by me.

MiYanni (approved):

src/dotnet-core-uninstall/Shared/Commands/UninstallCommandExec.cs
edvilme: Yes, this was briefly discussed with @joeloff https://github.com/dotnet/cli-lab/pull/331#discussion_r1974555062

PR #346, opened 2025-03-17
macOS: Fix corrupted binary

### LINKED ISSUE #344: [NETE2ESDK]The new uninstall tool could not run on Mac
**OS:** Mac intel and Arm64

**Reproduced steps:** 
1.  Download the tar.gz file and extract the tarball to a directory
2.  Run './dotnet-core-uninstall -h' and observe

**Expected Result:**
Uninstall tool can be worked normally. Take v.1.7.550802 for example

Image

**Actual Result:**
Running uninstall tool failed
On Mac intel
Image

On Mac M2
Image

edvilme:
Fixes #344 
macOS Binaries are getting corrupted after signing


edvilme:
Might be related to missing entitlements on entitlements.plist for runtime for codesign
https://github.com/dotnet/sdk/issues/24337

.vsts-ci.yml
```diff
@@ -130,20 +130,16 @@ extends:
               matrix:
                 ARM64:
                   _RID: osx-arm64
-                  _BuildConfig: Release
-                  _SignType: real
                   _DotNetPublishToBlobFeed: false
                 X64:
                   _RID: osx-x64
-                  _BuildConfig: Release
-                  _SignType: real
```
Forgind, line 139: We don't have to specify this anymore?

src/redist/targets/MacEntitlements/AddMacEntitlements.targets
```diff
@@ -0,0 +1,7 @@
+<Project>
+    <!-- Entitlements are needed but not automatically added for macOS. See https://github.com/dotnet/runtime/issues/113707 -->
+    <!-- This generates an ad-hoc signature that will later be resigned, but keeps the entitlements. -->
+    <Target Name="AddEntitlements" AfterTargets="GenerateLayout">
+        <Exec Command="codesign -s - -f --entitlements $(MSBuildThisFileDirectory)entitlements.plist $(ArtifactsDir)layout/dotnet-core-uninstall/dotnet-core-uninstall" />
```
Forgind, line 5: akoeplinger's version also had the `-` by itself, so I'm assuming it's correct, but it looked weird to me. What does this mean exactly?

Forgind (approved):

.vsts-ci.yml
edvilme: No need on adding it as a variable if it doesn't change in the matrix strategy. Better just adding it directly when calling the script

src/redist/targets/MacEntitlements/AddMacEntitlements.targets
edvilme: It refers to an empty certificate name. This is an ad-hoc signing, and it will be signed again, so I think it wouldn't make much difference

PR #382, opened 2025-03-31
CI check signatures

edvilme:
Signatures are not working properly after latest Arcade update.

This fixes the issues and adds an extra step for signature verification so that these are easier to debug in the future.

### DIFF
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -33,7 +33,6 @@ extends:
       binskim:
         analyzeTargetGlob: $(Build.SourcesDirectory)/artifacts/bin/**.dll;$(Build.SourcesDirectory)/artifacts/bin/**.exe;
     pool:
-      pool:
       name: $(DncEngInternalBuildPool)
       image: 1es-windows-2022
       os: windows
--- a/.vsts-ci.yml
+++ b/.vsts-ci.yml
@@ -110,14 +110,15 @@ extends:
               inputs:
                 sourceFolder: 'artifacts/packages/$(_BuildConfig)/Shipping/'
                 contents: '*.msi'
-                targetFolder: '$(Build.ArtifactStagingDirectory)'
+                targetFolder: '$(Build.ArtifactStagingDirectory)\artifacts'
             - task: MicroBuildCodesignVerify@3
               inputs:
-                TargetFolders: '$(Build.ArtifactStagingDirectory)'
+                TargetFolders: '$(Build.ArtifactStagingDirectory)\artifacts'
+                ExcludeFolders: '.git MicroBuild'
             - task: 1ES.PublishBuildArtifacts@1
               condition: and(eq(variables['system.pullrequest.isfork'], false), eq(variables['_BuildConfig'], 'Release'))
               inputs:
-                PathtoPublish: '$(Build.ArtifactStagingDirectory)'
+                PathtoPublish: '$(Build.ArtifactStagingDirectory)\artifacts'
                 ArtifactName: 'drop-windows'
                 publishLocation: 'Container'
                 parallel: true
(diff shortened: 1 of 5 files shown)


src/redist/targets/MacEntitlements/AddMacEntitlements.targets
```diff
@@ -1,7 +1,12 @@
 <Project>
     <!-- Entitlements are needed but not automatically added for macOS. See https://github.com/dotnet/runtime/issues/113707 -->
     <!-- This generates an ad-hoc signature that will later be resigned, but keeps the entitlements. -->
-    <Target Name="AddMacEntitlements" AfterTargets="GenerateLayout">
+    <ItemGroup>
+        <PackageReference Include="Microsoft.VisualStudioEng.MicroBuild.Core" Version="$(MicrosoftVisualStudioEngMicroBuildCoreVersion)" />
+    </ItemGroup>
+    <Target Name="AddMacEntitlements" 
+        BeforeTargets="SignFiles" 
+        AfterTargets="GenerateLayout">
```
edvilme, line 9: Important so that this runs _before_ MicroBuild signing

eng/SignVerifyIgnore.txt
edvilme, line 1: Needed for signing verification on windows

.vsts-ci.yml
```diff
@@ -112,11 +111,16 @@ extends:
               inputs:
                 sourceFolder: 'artifacts/packages/$(_BuildConfig)/Shipping/'
                 contents: '*.msi'
-                targetFolder: '$(Build.ArtifactStagingDirectory)'
+                targetFolder: '$(Build.ArtifactStagingDirectory)\artifacts'
```
edvilme, line 114: Added artifacts dir to isolate from other files when doing signing verification in the next step

src/redist/redist.csproj
```diff
@@ -9,7 +9,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\dotnet-core-uninstall\dotnet-core-uninstall.csproj" />
-    <PackageReference Include="Microsoft.DotNet.Build.Tasks.Installers" Version="10.0.0-beta.25124.4" Condition="'$(RID)' == 'win-x86'"/>
+    <PackageReference Include="Microsoft.DotNet.Build.Tasks.Installers" Version="10.0.0-beta.25124.4" />
```
MiYanni, line 12: Is the condition no longer relevant?

MiYanni (approved):

.vsts-ci.yml
```diff
@@ -99,8 +100,6 @@ extends:
               - name: _OfficialBuildArgs
                 value: ''
             steps:
-            - checkout: self
-              clean: true
```
nagilson, line 103: Did you remove clean: true because this is true by default?

.vsts-ci.yml
```diff
@@ -112,11 +111,16 @@ extends:
               inputs:
                 sourceFolder: 'artifacts/packages/$(_BuildConfig)/Shipping/'
                 contents: '*.msi'
-                targetFolder: '$(Build.ArtifactStagingDirectory)'
+                targetFolder: '$(Build.ArtifactStagingDirectory)\artifacts'
+            - task: MicroBuildCodesignVerify@3
+              inputs:
+                TargetFolders: '$(Build.ArtifactStagingDirectory)\artifacts'
+                ExcludeSNVerify: true
```
nagilson, line 118: What does ExcludeSNVerify do? May you provide the link to where this task is documented, I couldn't find it on the web.

nagilson (approved):
Please see my comments below, but looks good!

.vsts-ci.yml
edvilme: I discussed with @ mmitche offline, and according to him:
> It's because microbuild's tasks are getting installed to the source directory due to an arcade change to enable linux/mac signing. The checkout step you all run does a clean which deletes the files. 
> The thing is that the checkout step is not necessary since it happens by default IIRC

.vsts-ci.yml
edvilme: Sure: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/653/Codesign-Verification

src/redist/redist.csproj
edvilme: Oops you're right, missed that :)

PR #383, opened 2025-04-02
Mac: Add rid to tar.gz artifacts

edvilme:
Currently, the pipelines generates .tar.gz for both amr64 and x64 macs with the same name. This makes it so that they are identifiable by RID


Forgind (approved):
Is there automation that untars this automatically? It would feel like an important case for the SDK, but I don't think there would be many people in that case for this, so I think it's ok?

edvilme:
I don't think there is, as the next step in the pipeline is simply publishing the artifacts

PR #385, opened 2025-04-15
Update to xunit.v3 and Microsoft.Testing.Platform

Youssef1313:


marcpopMSFT:
Looks like two classes of changes. One is that the test assemblies all need to be exes. That should be fine here but if we move to v3 in other repos, we'll have to double check no one is assuming the app type.

The second is providing additional info in the custom facts. This appears to line up with the guidance in https://xunit.net/xunit.analyzers/rules/xUnit3003

Youssef1313:
> but if we move to v3 in other repos, we'll have to double check no one is assuming the app type.

If a a project forgets to set OutputType exe, there should be an error saying that.

Youssef1313:
> The second is providing additional info in the custom facts. This appears to line up with the guidance in https://xunit.net/xunit.analyzers/rules/xUnit3003

Yup. The caller info provided through attribute constructor is used mostly when running in Test Explorer. This avoids the hassle of how Test Explorer should calculate the file and line where a test is declared. Historically this was by reading PDB, then Roslyn exposed a helper to Test Explorer that does most a syntactic search (thus, it can fail in inheritance scenarios). We found it most ideal to do via caller info.

Youssef1313:
@marcpopMSFT Is it good merge then? 🎉

PR #377, opened 2025-03-25
[Bootstrapper] Add new project

edvilme:
Addresses #347 

----

_Also removes `MSBuildBinLogQuery`_

PR #386, opened 2025-04-15
Add a basic project and E2E test for the bootstrapper

### LINKED ISSUE #347: [Bootstrapper] Create the Project and Build in Pipeline
# 📓  Summary

Create a new folder for the bootstrapper that contains the main project for the bootstrapper. It should be Aot. Update the build pipelines and GitHub repository to have its own set of checks for building the bootstrapper. Also, add a single test which can do nothing except return success and add this as part of the pipeline check as well.

# Additional Context

The folder could be called 'bootstrapper' for now. Some additional properties you'll likely want in the project file:

TargetFramework - .NET 9 is the latest in support version of .NET.​

SelfContained – This tool is to help install .NET, so it shouldn't be required.​

RunAOTCompilation – WASM should be compiled AOT for Performance. Seems to be the property for WASM (as oppossed to PublishAot, and PublishReadyToRun which is the verison which allows reflection but has less perf gain.)​

WasmStripILAfterAOT – Remove IL files to reduce application size.​

PublishTrimmed – For Trimming Unused Framework Assemblies​

PublishSingleFile – We only want to publish a single binary.​

AssemblyName – DotnetSDKBootstrapper ? (Tentative) Don't include for now, probably.​

IsPackable – False, this is not meant to be a package.

nagilson:
Resolves https://github.com/dotnet/cli-lab/issues/347

- Adds 'dotnet-bootstrapper' as a project, and a test project.
- Utilizes the properties we'd expect to see from it in the final product.
- The name is still TBD, but don't block the PR on this.
- Contains a very simple E2E test to make sure the bootstrapper exits with code 0.
- Follows the existing repo conventions.
- Build and tests are already added, because the pipeline simply scans the folders (pretty nice.)
- The readme.md can be updated later.
- The build and tests can be separated/isolated later, and there is an issue for that.
- Sorry as I didn't see https://github.com/dotnet/cli-lab/pull/377 already existed. I think that wasn't finished though, and `MSBuildBinLogQuery` can be removed in a different PR.


nagilson:
I will be away but you are good to merge it if it is approved.

src/dotnet-bootstrapper/LocalizableStrings.Designer.cs
edvilme, line 1: Is this supposed to be here?

edvilme (approved):

src/dotnet-bootstrapper/LocalizableStrings.Designer.cs
nagilson: Yeah, its an empty file for when we add strings to this project

MiYanni: > Is this supposed to be here?

No, you shouldn't commit a .Designer file. It is generated on-build. You only need the .resx file.

src/dotnet-bootstrapper/dotnet-bootstrapper.csproj
```diff
@@ -0,0 +1,45 @@
...
+    <SelfContained>true</SelfContained>
+    <SignAssembly>true</SignAssembly>
+    <TargetFramework>net8.0</TargetFramework>
+    <RunAOTCompilation>true</RunAOTCompilation>
+    <PublishTrimmed>true</PublishTrimmed>
+    <IsPackable>false</IsPackable>
+    <PublishSingleFile>true</PublishSingleFile>
+    <PublishAot>true</PublishAot>
+    <WasmStripILAfterAOT>true</WasmStripILAfterAOT>
+    <RollForward>LatestMajor</RollForward>
```
dsplaisted, line 14: We should review which of these properties should be set.  I don't think `RollForward` applies to a self-contained app, and I don't think `WasmStripILAfterAOT` applies to non-WASM projects.  Some of the other AOT / publish properties may be redundant, I think the main one you need to specify is `PublishAOT`.

test/dotnet-bootstrapper.Tests/EndToEndTests.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Runtime.InteropServices;
+using Xunit;
+
+namespace Microsoft.DotNet.Tools.Bootstrapper
+{
+    public class EndToEndBootstrapperTest
+    {
+        private readonly static string artifactsDirectory = Path.GetFullPath(Path.Combine(Directory.GetCurrentDirectory(), "..", "..", "..", "..", "..", "artifacts"));
+        private readonly static string executablePath = Path.Combine(artifactsDirectory, "bin", "dotnet-bootstrapper", TestUtilities.GetConfiguration(), TestUtilities.GetTargetFramework(), TestUtilities.GetRuntimeIdentifier(),
+            RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "dotnet-bootstrapper.exe" : "dotnet-bootstrapper");
```
dsplaisted, line 18: I think it would be good to be a bit more resilient to changes in the folder layout.  Probably a good strategy is to find the repo root by looking for a `.git` folder and then use a relative path from there.

See `TestContext.GetRepoRoot` in dotnet/sdk for a sample of how to do that.  There's also a bunch more code in there for finding different folders, we probably don't need most of that complexity for this repo.

Issue #354, opened 2025-03-19
[Bootstrapper] Add logic for Global.Json


nagilson:
# 📓 Summary

Global.json parsing should be added to install the SDK. We don't have a great idea of how this should work and what it should do yet. Or if we should auto install etc based on the global.json and how the user interacts with this.

# Additional Context

[Supporting local SDK deployment in global.json by jaredpar · Pull Request #303 · dotnet/designs](https://github.com/dotnet/designs/pull/303) We may want to utilize this design. 

Consider that global.json may be in upper directories or child directories and how it should or shouldn't be applied.

# Blocked Until These Issues Are Done:

PR #391, opened 2025-04-23
[Bootstrapper] Fetch information from dotnet feeds

edvilme:
dotnet/sdk#50601 

Fetch information from dotnet feeds for search and install command. 
Added logic to find global.json file and fetch version information. 
Added logic for downloading files. 

## Search
Added search command to list all available sdk installs.
```
bootstrapper search [<channel>] [--allow-previews]
```
If no channel is provided, it will list all of them by default. 

## Install 
Added install command to install all assets from a specified version. 
```
bootstrapper install [<version>] [--allow-previews]
```
If no version is provided, it will look up the nearest global.json file and install the version specified there.

## Usage example

image


src/dotnet-bootstrapper/Commands/Install/InstallCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        Console.WriteLine($"Installing {product.ProductName} {latestRelease.Version}...");
+
+        foreach (ReleaseComponent component in latestRelease.Components)
+        {
+            Console.WriteLine($"Installing {component.Name} {component.DisplayVersion}...");
+
+            ReleaseFile releaseFile = component.Files.FirstOrDefault(file => file.Rid.Equals(Environment.OSVersion.Platform.ToString(), StringComparison.OrdinalIgnoreCase));
+            
+            releaseFile.DownloadAsync(Path.Combine(path, ".net", component.Name)).Wait();
```
edvilme, line 49: TODO: Logic for figuring out where to install to

src/dotnet-bootstrapper/Commands/Search/SearchCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace Microsoft.DotNet.Tools.Bootstrapper.Commands.Search;
+
+internal class SearchCommand(
+    ParseResult parseResult) : CommandBase(parseResult)
+{
+    private string _channel = parseResult.ValueForArgument(SearchCommandParser.ChannelArgument);
+    private bool _allowPreviews = parseResult.ValueForOption(SearchCommandParser.AllowPreviews);
+    public override int Execute()
+    {
+        List<Product> productCollection = [.. ProductCollection.GetAsync().Result];
```
joeloff, line 19: The releases.json only changes about once a month. You should consider caching the file on disk and only update it if there's a new version available.

baronfel: We should be using the same etag-based cache invalidation system that @nagilson has for the VSCode extension. ETags are _the_ way to handle cache invalidation of HTTP-delivered resources, and we shouldn't be reinventing the wheel.

src/dotnet-bootstrapper/BootstrapperUtilities.cs
```diff
@@ -0,0 +1,72 @@
...
+                throw new FileNotFoundException("No global.json file found in the directory tree.");
+            }
+            string directoryPath = Path.GetDirectoryName(globalJsonPath);
+            if (directoryPath == null)
+            {
+                throw new DirectoryNotFoundException("Directory path is null.");
+            }
+
+            // TODO: Replace with the actual installation directory.
+            return Path.Combine(directoryPath, ".dotnet.local");
```
edvilme, line 69: Where should we install to? Should we install to the `.dotnet` directory?

src/dotnet-bootstrapper/Commands/Search/SearchCommand.cs
```diff
@@ -43,10 +43,15 @@ public override int Execute()
...
             foreach (ProductRelease release in releases)
             {
+                // Get release.Sdks latest version
+                var latestSdk = release.Sdks
+                    .OrderByDescending(sdk => sdk.Version)
+                    .FirstOrDefault();
+
                 productMetadataTable.AddRow(
                     release.Version.ToString(), 
                     release.ReleaseDate.ToString("yyyy-MM-dd"),
```
joeloff, line 53: This forces the US approach for writing dates. Output, even from the CLI should be considered as a UI I think and use the UI thread culture for date formatting. This would get flagged under globalization during 1CS. Consider using the shortformat ("d") specifier. You should be able to do something like `ToString("d", CultureInfo.CurrentUICulture)`

src/dotnet-bootstrapper/Commands/Install/InstallCommand.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        return 0;
+    }
+
+    private static void DownloadAndExtractReleaseComponentFiles(ReleaseComponent component, string basePath)
+    {
+        if (component is WindowsDesktopReleaseComponent && !OperatingSystem.IsWindows())
+        {
+            return;
```
nagilson, line 68: Good call and edge case detection here 👌

nagilson: Probably also worth writing a test for.

src/dotnet-bootstrapper/Commands/Install/InstallCommand.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        ReleaseFile releaseFile = component.Files.FirstOrDefault(file =>
+            file.Rid.Equals(BootstrapperUtilities.GetRID(), StringComparison.OrdinalIgnoreCase) && (file.Name.EndsWith(".zip") || file.Name.EndsWith(".tar.gz")));
+
+        if (string.IsNullOrEmpty(releaseFile?.FileName))
+        {
+            Console.WriteLine($"\tNo suitable file found for {component.Name}");
```
nagilson, line 76: Nit: might be helpful to dump the component files array in the error.

src/dotnet-bootstrapper/Commands/Install/InstallCommand.cs
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    private static void DownloadAndExtractReleaseComponentFiles(ReleaseComponent component, string basePath)
+    {
+        if (component is WindowsDesktopReleaseComponent && !OperatingSystem.IsWindows())
+        {
+            return;
+        }
+
+        ReleaseFile releaseFile = component.Files.FirstOrDefault(file =>
```
nagilson, line 71:
```suggestion
        ReleaseFile releaseArchiveFile = component.Files.FirstOrDefault(file =>
```

nit: naming of the variable could specify that it is a file that represents just the component that has the zip or tar.gz, instead of any part of the release file. If I understand the code correctly 🤔

src/dotnet-bootstrapper/Commands/Install/InstallCommand.cs
```diff
@@ -0,0 +1,105 @@
...
+            file.Rid.Equals(BootstrapperUtilities.GetRID(), StringComparison.OrdinalIgnoreCase) && (file.Name.EndsWith(".zip") || file.Name.EndsWith(".tar.gz")));
+
+        if (string.IsNullOrEmpty(releaseFile?.FileName))
+        {
+            Console.WriteLine($"\tNo suitable file found for {component.Name}");
+            return;
+        }
+
+        string zipPath = Path.Combine(basePath, releaseFile.FileName);
```
nagilson, line 80:
```suggestion
        string archiveInstallPath = Path.Combine(basePath, releaseFile.FileName);
```
Nit: Update variable name to be os-agnostic and indicate that it is something that does not yet exist.

src/dotnet-bootstrapper/Commands/Install/InstallCommand.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        if (File.Exists(zipPath))
+        {
+            Console.WriteLine($"\t{component.Name} already exists at {zipPath}");
+            return;
```
nagilson, line 85: I don't think we should return if the zip already exists, as this could allow someone to get into a 'permabad' state. Instead, we should skip the download of the zip but still extract it, as long as it looks like a valid zip (maybe check the sha / file size)

src/dotnet-bootstrapper/Commands/Install/InstallCommand.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        try
+        {
+            releaseFile.DownloadAsync(zipPath)?.Wait();
+
+            // Extract the downloaded file
+            ZipFile.ExtractToDirectory(zipPath, Path.ChangeExtension(zipPath, ""));
```
nagilson, line 93: It would be ideal to check whether it has already been extracted to this directory before so we can skip this labor, but iff the extraction resulted in a success and valid installation.

src/dotnet-bootstrapper/Commands/Install/InstallCommand.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            Console.WriteLine($"\tExtracted {component.Name} to {Path.ChangeExtension(zipPath, "")}");
+
+            // Delete the downloaded file
+            File.Delete(zipPath);
```
nagilson, line 98: We should probably have an issue w.r.t caching this zip file in case someone installs again? Not very high priority and definitely not a blocker.

src/dotnet-bootstrapper/Commands/Install/InstallCommand.cs
nagilson, line 93: I see that DownloadAsync verifies the hash. It's a bit awkward though in that it will do a download even if the dest file exists and always replace it. Ideally we could verify the hash of an existing file and re-utilize it, though I know that's not as relevant in the current context, since we delete the file every time.

nagilson, line 93: One thing to consider: What if someone runs multiple bootstrapper commands at the same time? Will the directory extraction clobber over each other? Should there be some sort of lock on this operation that can be shared across processes? 

What if someone pkills the bootstrapper as it is in the middle of extracting the directory? Then it will leave the directory in a bad state, I believe. We may want to consider cleaning the directory up beforehand (admittedly, this would be bad if someone incorrectly configured their path and then it wiped something else out) or handling these scenarios better with a different approach.

I also don't know if I would block on this but I do consider it higher priority after this.

nagilson, line 93: It's likely worth doing a dive into each of the possible error cases here to handle them: PathTooLong (does this function respect the OS setting or rely on the default?) -> Directory Not Found - Do we need to create the directory? Unauthorized -> Should we try a fallback directory (probably not for now)

nagilson (commented):
Thank you for writing this up, I'm excited to see a lot of progress here!

In the very early days of a project, it's very hard to say what should/shouldnt go in a separate PR and what the 'bar' should be. Some of what I mention below I do think are things we don't need to block this PR on (I mostly remark when that is the case), though some of them are. I think this is a great start 🔥 and it is normal to have a lot of feedback and things to consider.

There is a balance of 'moving fast' vs carefully considering every possible edge case. Since we are not even a launched product I would say we are on the 'moving fast' side for now, to some extent. I would defer to @baronfel and @marcpopMSFT for a broader priority discussion on some of these items.

src/dotnet-bootstrapper/BootstrapperUtilities.cs
```diff
@@ -0,0 +1,72 @@
...
+                .ToString();
+
+            ReleaseVersion version = ReleaseVersion.Parse(sdkVersion);
+            Console.WriteLine($"Found version {version.Major}.0 in global.json");
+            return $"{version.Major}.0";
+        }
+
+        public static string GetInstallationDirectoryPath()
+        {
+            string globalJsonPath = GlobalJsonUtilities.GetNearestGlobalJsonPath(Environment.CurrentDirectory);
```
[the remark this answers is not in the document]
nagilson: I think we can defer this to https://github.com/dotnet/cli-lab/issues/354.

src/dotnet-bootstrapper/BootstrapperUtilities.cs
```diff
@@ -0,0 +1,72 @@
...
+            if (globalJsonPath == null)
+            {
+                throw new FileNotFoundException("No global.json file found in the directory tree.");
```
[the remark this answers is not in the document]
nagilson: https://github.com/dotnet/cli-lab/issues/354

src/dotnet-bootstrapper/Commands/Search/SearchCommand.cs
edvilme: From what I've seen, I don't think that it does. I do agree that it would make a lot of sense to implement it there

src/dotnet-bootstrapper/GlobalJsonUtilities.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.IO;
+using System.Text.Json;
+
+namespace Microsoft.DotNet.Tools.Bootstrapper
+{
+    internal static class GlobalJsonUtilities
+    {
+        public static string GetNearestGlobalJsonPath(string basePath)
```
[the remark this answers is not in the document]
nagilson: > This is good for now, I think the runtime implemented some global json logic and it loks like CliTemplateEngineHost also did. We should probably try to share that implementation in the future, and I can create an issue for that.

https://github.com/dotnet/cli-lab/issues/393

nagilson:
Great work updating things in response to my feedback above; if there's anything I can clarify then please let me know, happy to chat! 😊

marcpopMSFT:
Note: Long term effort that needs reevaluation for our backlog.

lextm:
Fully understand that the usefulness of metadata from .NET feeds but maybe dotnet-core-uninstall should remain as simple as possible (no network connectivity required).

In case end users do need to see more information about the versions they installed, a wrapper tool can combine both the metadata and dotnet-core-uninstall output and provide a streamlined view. [DotUninstall](https://github.com/lextudio/DotUninstall) is an example.

